PR #4, opened 2015-07-12
add feature to create Capture from a saved .pcap file

darxriggs:
* new method `Capture::from_file`
* renamed method `Capture::new` to `Capture::from_device`


ebfull:
Great! Thank you.

PR #14, opened 2015-08-05
make Packet public

AAG81:
Hi
I need caplen and len so I made header and data field public (in Packet Struct)


ebfull:
I've decided instead to expose the header via a new struct which would use libc's types instead. I've re-calibrated that API in [1ac837d](https://github.com/ebfull/pcap/commit/1ac837d7a3986ddf44bf7cad910b100e397957cc). Let me know if those changes suffice for you, for now I'm closing this PR.

PR #17, opened 2015-09-06
Debug for PacketHeader

darxriggs:


darxriggs:
Using [Formatter.debug_struct](https://doc.rust-lang.org/core/fmt/struct.Formatter.html#method.debug_struct) would be another option:

    fmt.debug_struct("PacketHeader")
        .field("ts", &(format!("{}.{:06}", self.ts.tv_sec, self.ts.tv_usec)))
        .field("caplen", &self.caplen)
        .field("len", &self.len)
        .finish()

src/raw.rs
```diff
@@ -589,6 +589,13 @@ pub struct PacketHeader {
     pub len: u32,
 }
 
+impl ::std::fmt::Debug for PacketHeader {
+    fn fmt(&self, f: &mut ::std::fmt::Formatter) -> ::std::fmt::Result {
+        write!(f, "PacketHeader {{ ts: {}.{:06}, caplen: {}, len: {} }}",
```
darxriggs: Using [Formatter.debug_struct](https://doc.rust-lang.org/core/fmt/struct.Formatter.html#method.debug_struct) would be another option:

    f.debug_struct("PacketHeader")
        .field("ts", &(format!("{}.{:06}", self.ts.tv_sec, self.ts.tv_usec)))
        .field("caplen", &self.caplen)
        .field("len", &self.len)
        .finish()

ebfull:
Thank you, perfect. :)

PR #22, opened 2015-09-09
Switched from pcap_inject to pcap_sendpacket to fix #21.

### LINKED ISSUE #21: Could you please add support to pcap_sendpacket for winpcap?
Winpcap does export function pcap_sendpacket and I saw the pcap_sendpacket function has been in raw.rs (commented recently), is there a method merging it to the function sendpacket in lib.rs to support sending packets for winpcap?

Thanks.

ebfull:
cc @ZhangQin3

Closes #21.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ documentation = "https://ebfull.github.io/pcap/"
 license = "MIT"
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


ebfull:
I tested that this at least links with winpcap but I haven't tested on Windows.

PR #37, opened 2016-02-25
Implement AsRawFd for Capture for pcap_fileno ffi

hjr3:
The `AsRawFd` trait does not return a Result. However, the Capture
struct tracks the state of the capture. According to man, `pcap_fileno`
will should only return -1 for an offline/inactive capture. Thus, we
should never panic.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,6 +59,7 @@ use std::mem::transmute;
 use std::str;
 use std::fmt;
 use self::Error::*;
+use std::os::unix::io::{RawFd, AsRawFd};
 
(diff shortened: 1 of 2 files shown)


hjr3:
@ebfull I just made a decision on how `pcap_fileno` should work. If you feel differently, please let me know. I am happy to change my approach.

Also, I am not sure if there should be a test for this?

Thank you for starting this crate!

ebfull:
It may be useful to implement for `FromRawFd` and `IntoRawFd` as well.

ebfull:
... but this works for now though, thanks!

PR #39, opened 2016-03-07
Conditional compile unix specific features for Windows users.

kestein:
This patch adds a conditional compilation to the RawFd and AsRawFd
features that are only available in Unix.


ebfull:
Perfect! Thanks!

PR #40, opened 2016-03-17
Add direction method

aleksander:


ebfull:
This looks fine to me, I don't know too much about this API though.

ebfull:

PR #41, opened 2016-03-17
Add nanosecond precision methods

aleksander:


src/lib.rs
```diff
@@ -348,6 +348,24 @@ impl Capture<Offline> {
             })
         }
     }
+
+    /// Opens an offline capture handle from a pcap dump file, given a path.
```
ebfull: This should specify the point of the second function so that people reading the documentation can understand the difference.

src/lib.rs
```diff
@@ -423,6 +441,13 @@ impl Capture<Inactive> {
         }
     }
 
+    pub fn nano(self) -> Capture<Inactive> {
```
ebfull: This should allow the caller to enable or disable the precision instead of always enabling it.

src/lib.rs
aleksander: Fixed.

ebfull:
Thanks!

ebfull:

PR #44, opened 2016-04-11
Add a 'Dead' capture type 

aphistic:
This will allow creating pcap save files without needing to capture from an actual interface.

It was a pretty straightforward implementation but there were two things I wasn't sure how you wanted to handle.

I named the function to create them "dead" because I couldn't think of anything better since it's not really "from_" anything.
I wanted to have a sane default for the snaplen parameter but I didn't want to end up with a bunch of functions to create the struct, so I defaulted it to 65535.  I figured if anyone wants to change it they can use snaplen().

Let me know if you want me to change any of this! I think this will work for me as I was able to successfully create a new capture file with the code.  I'm not sure I'll get my "write directly to a Vec<u8>" because of how pcap writes the packets but I can always throw it in a temp file.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -715,26 +715,6 @@ impl Capture<Dead> {
             })
         }
     }
-
-    /// Creates a "fake" capture handle for the given link type.  Takes an additional
-    /// precision argument specifying the time stamp precision desired.
-    pub fn dead_with_precision(linktype: Linktype, precision: Precision) -> Result<Capture<Dead>, Error> {
-        unsafe {
-            let handle = raw::pcap_open_dead_with_tstamp_precision(linktype.0, 65535, match precision {
-                Precision::Micro => 0,
-                Precision::Nano => 1,
-            });
-
-            if handle.is_null() {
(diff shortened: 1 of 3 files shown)


ebfull:
Excellent. The only thing I would like to see is a test added, since this shouldn't (hopefully) require privileges to execute, since it's not attaching to a network interface.

aphistic:
Ah, sure. I'll work on adding that.

aphistic:
It looks like the Ubuntu that travis is running (12.04LTS) has a super-old version of libpcap on it (1.1.1) that doesn't have the precision version of pcap_open_dead(). Which version of libpcap are you targeting?  Should I just remove that function?

ebfull:
I would just remove the function, yeah. Is there not another way to configure the precision manually anyway?

aphistic:
Not that I saw in the header file of the 1.1 branch.  I just removed it completely for now.

ebfull:
Great job! I believe this interface is available on Windows so we hopefully won't have any trouble there.

ebfull:

PR #46, opened 2016-04-19
Allow creation of Savefiles in-memory only

aphistic:
This adds a "savememory()" function to Capture that uses shared memory to write the pcap file directly to memory instead of requiring a file on the filesystem first.

At this point it doesn't support Windows and OS X.  I tried to get it running on OS X but apparently it really doesn't like fseek (or much of anything) on shared memory so I left that as an exercise to someone who's better versed in Windows and/or OS X.

I screwed up my rebase so there are some double commits and a merge, sorry. =/

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 libc = "0.2.7"
+rand = "0.3.14"
 
 [lib]
 name = "pcap"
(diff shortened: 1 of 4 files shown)


ebfull:
This is cool; I could see the use case. :+1: 

I'd be happy to review this but it would be easier if it were all squashed into a single commit and rebased on top of the current master.

aphistic:
Closing this PR because it's easier to just create a new one than fix my weird commit history... That's the last time I forget to update from upstream before I start work! :)

PR #47, opened 2016-04-19
Add savememory() to allow Savefiles to be created in memory

aphistic:
This adds a "savememory()" function to Capture that uses shared memory to write the pcap file directly to memory instead of requiring a file on the filesystem first.

At this point it doesn't support Windows and OS X. I tried to get it running on OS X but apparently it really doesn't like fseek (or much of anything) on shared memory so I left that as an exercise to someone who's better versed in Windows and/or OS X.

This is a new version of #46 because I screwed that one up. =/

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,8 +47,11 @@
 //! ```
 
 extern crate libc;
+extern crate rand;
 
 use unique::Unique;
+#[cfg(not(any(windows, target_os="macos")))]
+use rand::Rng;
 use std::marker::PhantomData;
 use std::ptr;
 use std::ffi::{CStr,CString};
@@ -71,7 +74,7 @@ const PCAP_ERROR_NOT_ACTIVATED: i32 = -3;
 const PCAP_ERRBUF_SIZE: usize = 256;
 
 /// An error received from pcap
-#[derive(Debug)]
+#[derive(Debug, PartialEq)]
 pub enum Error {
     MalformedError(str::Utf8Error),
     InvalidString,
@@ -80,6 +83,7 @@ pub enum Error {
     TimeoutExpired,
     NoMorePackets,
     InsufficientMemory,
+    General(String),
 }
 
 impl Error {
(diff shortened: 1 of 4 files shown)


ebfull:
I need to set aside time to review this, my apologies!

aphistic:
No problem! I'm just using my fork for development at the moment.

src/lib.rs
```diff
@@ -770,6 +786,118 @@ impl Drop for Savefile {
...
+
+#[cfg(not(any(windows, target_os="macos")))]
+impl Savememory {
+    /// Creates a new in-memory pcap file for the given pcap handle.
+    fn new(pcap_handle: *mut raw::pcap_t) -> Result<Savememory, Error> {
+        // Generate a random name so we don't collide
+        let mut name = vec![];
+        let mut name_rng = rand::thread_rng();
+        let mut name_gen = name_rng.gen_ascii_chars();
+        for _ in 0..20 {
```
ebfull: You should be able to do `for c in name_rng.gen_ascii_chars().take(20) { ... }`.

src/lib.rs
```diff
@@ -770,6 +786,118 @@ impl Drop for Savefile {
...
+            match name_gen.next() {
+                Some(c) => name.push(c as u8),
+                None => return Err(Error::InsufficientMemory),
```
ebfull: I don't understand this error, does `gen_ascii_chars` really halt for some memory exhaustion or something?

aldanor:
@aphistic While this is quite reasonable, it's (1) very Linux-specific, (2) represents functionality that's not present in the original libpcap. I believe a more general approach would be to provide friendly wrappers for `pcap_fopen_*()` API, for both reading/writing savefiles. I believe #66 addresses that, would it resolve this PR to some extent?

Wojtek242:
Cleaning up some old PRs. If you would like to see this code in pcap, please update and resubmit.

PR #52, opened 2016-05-03
allow pcap_next_ex() to return codes >= 1

### LINKED ISSUE #51: pcap_next_ex can return other positive return codes to indicate success (e.g., with NFLOG)
I'm trying to use pcap's nflog input to analyze iptables-filtered packets. Here's the entirety of the code I'm using:

```rust
extern crate pcap;

use pcap::Device;

fn main() {
    let device = pcap::Device { name: "nflog:40".to_string(), desc: None };
    let mut cap = device.open().expect("Cannot open device");

    while let Ok(packet) = cap.next() {
        println!("{:?}", packet);
    }
}
```

Then I added an iptables rule to capture all outbound traffic from my test machine:
```
sudo iptables -A OUTPUT -j NFLOG --nflog-group 40
```

The netlink socket gets established correctly, because after the first packet is sent, my Rust program dies with the following:
```
thread '<main>' panicked at 'internal error: entered unreachable code', /home/vagrant/.cargo/registry/src/github.com-88ac128001ac3a9a/pcap-0.5.7/src/lib.rs:639
```

I added a `println!` to print out the return code from `pcap_next_ex`, and it's returning `9`. Looking at the source code for my version of pcap (`1.4.0-4.20130826git2dbcaa1`, the Red Hat/CentOS 6 version), `pcap_next_ex` will return the result of `pcap_read()`, whose source code has the following comment:

```c
        /*
         * Return codes for pcap_read() are:
         *   -  0: timeout
         *   - -1: error
         *   - -2: loop was broken out of with pcap_breakloop()
         *   - >1: OK
         * The first one ('0') conflicts with the return code of 0 from
         * pcap_offline_read() meaning "end of file".
        */
```

This contradicts the RETURN VALUE section of the `pcap_next_ex` man page (which says that OK will only return 1), but I'm more inclined to believe the source  than the man page.

I applied the following patch to this library and it seems to work:

```diff
diff --git a/src/lib.rs b/src/lib.rs
index 03fd0ae..fce71e2 100644
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -613,7 +613,7 @@ impl<T: Activated + ?Sized> Capture<T> {
             let mut header: *mut raw::Struct_pcap_pkthdr = ptr::null_mut();
             let mut packet: *const libc::c_uchar = ptr::null();
             match raw::pcap_next_ex(*self.handle, &mut header, &mut packet) {
-                1 => {
+                i if i >= 1 => {
                     // packet was read without issue
                     Ok(Packet {
                         header: transmute(&*header),
```

Roguelazer:
Fixes #51

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -613,7 +613,7 @@ impl<T: Activated + ?Sized> Capture<T> {
             let mut header: *mut raw::Struct_pcap_pkthdr = ptr::null_mut();
             let mut packet: *const libc::c_uchar = ptr::null();
             match raw::pcap_next_ex(*self.handle, &mut header, &mut packet) {
-                1 => {
+                i if i >= 1 => {
                     // packet was read without issue
                     Ok(Packet {
                         header: transmute(&*header),
@@ -635,7 +635,7 @@ impl<T: Activated + ?Sized> Capture<T> {
                     Err(NoMorePackets)
                 },
                 _ => {
-                    // libpcap only defines codes 1, 0, -1, and -2
+                    // libpcap only defines codes >=1, 0, -1, and -2
                     unreachable!()
                 }
             }


ebfull:
Thanks!

PR #55, opened 2016-07-10
Support "immediate mode"

### LINKED ISSUE #45: Why no pcap_set_immediate_mode / pcap_setmintocopy ?
Hi! Congatulations on porting this to Rust, good work!

However as I was scanning your code to decide porting my app to Rust from C++, I was surprised by the absence of these two function calls.
Both of them serve to enable a low latency mode in which pcap/WinPcap reports packets to your application nearly instantly, without buffering them for efficiency.

This of course causes a lot more of context switches, but enables applications to serve as routers, firewalls or something else, where delays introduced by the buffering (Win)pcap  usually does is a pitfall.

I saw that `pcap_set_immediate_mode` is commented out in https://github.com/ebfull/pcap/blob/master/src/raw.rs#L675. Is there a reason for that? Or it was not necessary for you at the time, and decided to leave it out of your lib?

With respect to `pcap_setmintocopy` this is a WinPcap exclusive, as `pcap_set_immediate_mode` is not available in Windows. I don't know if it's simply a substitute or if more complex work is carried out by this function in order to get this feature on windows.

I look forward hearing from you.
Dark_eye

ryan-robeson:
This PR brings support for libpcap's "immediate mode" on Windows (setmintocopy), Mac(ioctl - BIOCIMMEDIATE), and Linux(set_immediate_mode).

It is very much a **WIP** right now. This is my first time playing with Rust or libpcap, so there are several things I'd like to clarify before calling this done. Also, I'm not familiar with proper Rust documentation yet, but I'll get that corrected ASAP. I just wanted to get this out there so we could get the discussion going.

I'll update this PR from feedback, and rebase it when it's done so you don't have a bunch of useless commits.

The questions I have are documented in the diff for simplicity (I hope). Please let me know if they'd be better in a comment.

Largely, I'm just hoping to clarify preferences. The biggest issue is that libpcap on OS X requires an activated capture to enable immediate mode, and libpcap on Linux requires an inactive capture. This causes an inconsistent API `.immediate_mode(true).open().unwrap()` vs `.open().unwrap().immediate_mode(true)`. 

It seems to me that `immediate_mode()` should be defined on `Capture<Inactive>` for every OS. For those that require an active capture, `immediate_mode() -> Capture<Inactive>` should set a flag that is checked for when the capture is `open()`'d. `open()` would then take the necessary steps to enable "immediate mode". This would keep the API consistent cross-platform. 

Is there a better way? If not, what is the best way to implement this?

Thanks
## Test program

``` rust
extern crate pcap;

fn main() {

    // immediate_mode(false): print packets after timeout or buffer fills
    // immediate_mode(true): print packets as soon as they arrive
    let mut cap = pcap::Capture::from_device(YOURDEVICENAMEHERE).unwrap()
      .timeout(10000) // When "immediate mode" is enabled, this tells you that no packets were received and keeps it from blocking on some systems
      // .immediate_mode(true).open().unwrap(); // Linux only: uncomment this line and comment the next.
      open().unwrap().immediate_mode(true); 

   // Wake-on-lan packets are an easy way to test immediate mode because they're
   // easy to trigger and fairly rare.
   // If you have "wakeonlan" installed, simply:
   // `wakeonlan $RANDOMMAC`
   // With immediate mode enabled, running that command multiple times should print packets 
   // right away.
   // Without it, it should take at least 10000ms (10s) for the first packet to print (Unless
   // you run `wakeonlan` in a loop and manage to fill up the buffer before the 10s timeout,
   // but even that will show a delay compared to immediate mode.
    cap.filter("ether proto 0x0842 or udp port 7 or udp port 9").unwrap();

    println!("{:?}", cap.next());
    println!("{:?}", cap.next());
    println!("{:?}", cap.next());
    println!("{:?}", cap.next());
    println!("{:?}", cap.next());
}
```

Closes #45

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -745,6 +745,38 @@ impl Capture<Active> {
             Ok(self)
         })
     }
+
+    /// Set immediate mode on or off for a not-yet-activated capture handle.
+    /// Separate implementations are required for Windows, Mac, and Linux.
+    // TODO: This may need to be on Capture<Inactive> for Linux. Not ideal.
+    //#[cfg(linux)]
+    //pub fn immediate_mode(self, to: bool) -> Capture<Inactive> {
+    //    unsafe {
+    //        raw::pcap_set_immediate_mode(*self.handle, if to {1} else {0});
+    //        self
+    //    }
+    //}
+
+    /// Set immediate mode on or off for a not-yet-activated capture handle.
+    /// Separate implementations are required for Windows, Mac, and Linux.
+    /// OSX doesn't support pcap_set_immediate.
+    // Idea: If Win and Linux are both set on Inactive Capture,
+    //       have the OS X method set a flag that is checked for when the
+    //       capture is activated which actually sets the immediate mode
+    //       in order to keep the API consistent.
+    // Question: Do BSD's use BIOCIMMEDIATE or pcap_set_immediate_mode?
+    //           This guard should be updated if the former.
+    #[cfg(target_os = "macos")]
+    pub fn immediate_mode(self, to: bool) -> Capture<Active> {
+        unsafe {
+            let i: libc::c_uint = if to {1} else {0};
+            // See /usr/include/sys/ioccom.h for the steps
+            // required to derive BIOCIMMEDIATE.
+            const BIOCIMMEDIATE: libc::c_ulong = 0x80044270;
+            libc::ioctl(raw::pcap_fileno(*self.handle), BIOCIMMEDIATE, &i);
+            self
+        }
+    }
 }
 
 impl Capture<Dead> {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -557,6 +557,32 @@ impl Capture<Inactive> {
         self
     }
 
+    /// Set immediate mode on or off for a not-yet-activated capture handle.
+    /// Separate implementations are required for Windows, Mac, and Linux.
+    /// Windows and Mac do not support pcap_set_immediate.
+    //
+    // Linux must be set on an Inactive Capture. Mac requires an Active Capture.
+    // Currently unsure if Windows cares (It definitely works with an Active Capture).
+    //
+    // Question 1:
+    //   Context: It's my understanding that determining the installed version of libpcap
+    //            is problematic. Thus, checking for feature support is difficult.
+    //   * Should we change this guard to Not Win + Not Mac and rely on Issues to find
+    //     find out where this fails?
+    //   * Or, should we whitelist OS's as they're needed? (I'm leaning toward the former)
+    //
+    // Question 2:
+    //   * pcap_set_immediate_mode returns PCAP_ERROR_ACTIVATED if called on an
+    //     activated capture handle. Since this is defined on Capture<Inactive>,
+    //     we can safely ignore this right?
+    #[cfg(target_os = "linux")]
+    pub fn immediate_mode(self, to: bool) -> Capture<Inactive> {
+        unsafe {
+            raw::pcap_set_immediate_mode(*self.handle, if to {1} else {0});
+            self
+        }
+    }
+
     /// Set the time stamp precision returned in captures.
     #[cfg(not(windows))]
     pub fn precision(self, precision: Precision) -> Capture<Inactive> {
@@ -749,21 +775,9 @@ impl Capture<Active> {
     /// Set immediate mode on or off for a not-yet-activated capture handle.
     /// Separate implementations are required for Windows, Mac, and Linux.
     /// Windows and Mac do not support pcap_set_immediate.
-    // TODO: This may need to be on Capture<Inactive> for Linux. Not ideal.
-    //#[cfg(linux)]
-    //pub fn immediate_mode(self, to: bool) -> Capture<Inactive> {
-    //    unsafe {
-    //        raw::pcap_set_immediate_mode(*self.handle, if to {1} else {0});
-    //        self
-    //    }
-    //}
-
-    /// Set immediate mode on or off for a not-yet-activated capture handle.
-    /// Separate implementations are required for Windows, Mac, and Linux.
-    /// Windows and Mac do not support pcap_set_immediate.
-    // Idea: If Linux is set on Inactive Capture,
+    // Idea: Since Linux is set on Inactive Capture, should we
     //       have the OS X method set a flag that is checked for when the
-    //       capture is activated which actually sets the immediate mode
+    //       capture is activated which actually enables immediate mode
     //       in order to keep the API consistent.
     // Question 1: Do BSD's use BIOCIMMEDIATE or pcap_set_immediate_mode?
     //             This guard should be updated if the former.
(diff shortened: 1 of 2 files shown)


ebfull:
Nice work!

I haven't had time to look at this, but I believe your approach to setting a flag for Mac to check on `open()` is the best approach. It will require having the `Capture<Inactive>` struct store that flag. The best way to accomplish that is to add an associated type to the State trait called `Metadata` (which is forced to implement `Default + Sized`) and adding a field to `Capture` like:

```rust
pub struct Capture<T: State + ?Sized> {
    handle: Unique<raw::pcap_t>,
    metadata: T::Metadata
}
```

For `Inactive` (on Mac) this `Metadata` type can contain the flag. This avoids forcing `Capture<Active>` or other instantiations to contain a boolean that isn't used.

ebfull:
Note that I don't think `_marker: PhantomData<T>` will be necessary after you add that field which depends on `T`. I don't know why though.

ryan-robeson:
Thanks for the feedback! I'll look into that

ryan-robeson:
I've not had much time to work on this, or on learning Rust in general unfortunately, but I haven't forgotten about it.

Question. From your comments and my interpretation of the Rust docs, I've tried to add the `Metadata` associated type to the `State` trait. 

```rust
pub unsafe trait State {
  type Metadata;
}

#[derive(Default)]
struct StateMetadata {
    immediate_mode_enabled: bool,
}

unsafe impl State for Inactive {
  type Metadata = StateMetadata;
}
```

From your response, it sounds like `State::Metadata` should not exist on anything but `Capture<Inactive>`.  However, the compiler complains (see below) that it must be specified on `Activated`, presumably because of the trait inheritance. 

So, have I gone about this the wrong way (highly likely given my inexperience with Rust)? If not, what's the best way to proceed? 

* Use the `StateMetadata` type as the `Metadata` type in `Activated`
* Use an empty struct or some other type to make the compiler happy
* Or something else entirely?

Thanks

## Compiler Error
```
src/lib.rs:850:49: 850:58 error: the value of the associated type `Metadata` (from the trait `State`) must be specified [E0191]
src/lib.rs:850 impl<T: Activated> From<Capture<T>> for Capture<Activated> {
                                                               ^~~~~~~~~
src/lib.rs:850:49: 850:58 help: run `rustc --explain E0191` to see a detailed explanation
src/lib.rs:851:41: 851:50 error: the value of the associated type `Metadata` (from the trait `State`) must be specified [E0191]
src/lib.rs:851     fn from(cap: Capture<T>) -> Capture<Activated> {
                                                       ^~~~~~~~~
src/lib.rs:851:41: 851:50 help: run `rustc --explain E0191` to see a detailed explanation
error: aborting due to 2 previous errors
error: Could not compile `pcap`.
```

ebfull:
Yes, use an empty tuple `()` (unit type) for the associated type on `Activated`. (No time to double-check but that's what I meant to say before.)

ryan-robeson:
Alright. I've been wrestling with this for a couple hours now (Just noticed your response). 

I did try implementing it with empty tuples, but the compiler kept throwing errors (which I now believe were unrelated), so I've got it compiling with `metadata = Option<StateMetadata>` at the moment. Setting `metadata: None` as appropriate.

I assumed these lines would fix the error above:

```rust
unsafe impl State for Active {
  type Metadata = Option<StateMetadata>;
}
unsafe impl State for Offline {
  type Metadata = Option<StateMetadata>;
}
unsafe impl State for Dead {
  type Metadata = Option<StateMetadata>;
}
```

But, they didn't. I also had to change the following:

```diff
-impl<T: Activated> From<Capture<T>> for Capture<Activated> {
-    fn from(cap: Capture<T>) -> Capture<Activated> {
+impl<T> From<Capture<T>> for Capture<Activated<Metadata=T::Metadata>> where T: Activated + State<Metadata=Option<StateMetadata>> {
+    fn from(cap: Capture<T>) -> Capture<Activated<Metadata=T::Metadata>> {
         unsafe { transmute(cap) }
     }
 }
```

It seems that Rust won't infer the associated types. Is there a trick to it, or is this expected behavior? 

Also, I'm happy to change back to tuples if you want. It would also probably make sense to revert `Option<StateMetadata>` to `StateMetadata` if we do that.

stbuehler:
I think the platform abstraction should be provided by libpcap, not by the rust wrapper. I.e. one should make `pcap_set_immediate_mode` work across all platforms, instead of building workarounds using different libpcap APIs in rust.

You claim:
> The biggest issue is that libpcap on OS X requires an activated capture to enable immediate mode, and libpcap on Linux requires an inactive capture.

I don't think that is true, at least not in the current version of libpcap; the man page for `pcap_set_immediate_mode` (and [implementation](https://github.com/the-tcpdump-group/libpcap/blob/master/pcap.c#L1408) too!) is very clear: it only works on inactive handles. Maybe it got fixed in the meantime :)

Afaics all platforms either support immediate mode or fail to activate the handle.

See https://github.com/stbuehler/pcap/commit/immediate-mode for what I'd do for - no magic at all. If there is interest in my patch I can submit it as PR (although it is rather trivial ofc).

ryan-robeson:
Which version of OS X are you looking at? Perhaps a more recent version of libpcap is included with 10.12? I was on 10.9.5 when I last looked at this. Other languages' wrappers seemed to take the same approach at the time.

I do agree that handling it in libpcap is ideal, and it's the way to go if recent versions of OS X are now including a libpcap that supports it (10.9 is pretty old now). 

The project I wanted to use this lib for is on the back burner for the foreseeable future, but I'm sure others would appreciate the PR for a working implementation (even if it's only a few lines now). 

Just to clarify, `pcap_set_immediate_mode` is supported on Windows now too?

fcangialosi:
Just wanted to leave one sample saying this works great for me out of the box (on Linux). Before adding this I was seeing a pesky extra ~4ms of delay that I needed to get rid of and this did the trick. 

Thanks for putting this together @ryan-robeson!

drrlvn:
@ryan-robeson Would it perhaps be possible to rebase your PR on top of master?

If this isn't merged then I could at least use your fork if it was rebased.

ryan-robeson:
@drrlvn, this is now rebased on master. I haven't tested it, but the changes are pretty small, so hopefully you won't have any problems with it. 

I also published a `set-immediate-mode-metadata` branch on my fork that contains the work I had started that is mentioned above. I tried rebasing it too, but got lost in a couple of the conflicts. Pretty sure it would require some changes to work with the current codebase, and I've not written any Rust or used this library in a couple years. It may be a starting point though if you need those changes.

Hope this helps!

drrlvn:
@ryan-robeson Thank you very much! It's a great help and I really appreciate you doing this even though you no longer use the crate or Rust.

ryan-robeson:
@drrlvn Happy to help!

drrlvn:
I don't really understand why but with libpcap 1.9.1 `capture.next()` waits for ~1 minute before delivering packets, but setting immediate mode using @ryan-robeson's branch solves this issue (needless to say I'm not using timeouts). I'm not sure I identified the problem correctly so if someone could correct me that would be great.

ryan-robeson:
@drrlvn, off the top of my head, IIRC pcap uses a buffer for incoming packets and doesn't start returning any until that buffer fills up (assuming immediate mode is off). Perhaps you are capturing a low volume of traffic which could take a minute to fill the buffer?

Just going from memory, haven't verified that

Wojtek242:
Obsoleted by #140 which makes use of the new build-time libpcap version detection to simplify things.

PR #57, opened 2016-09-10
allow appending to save files

maltek:
This adds a parameter to Capture::savefile() that allows appending to the pcap file instead of creating a new one.

If you'd prefer another function over a boolean parameter like in the underlying pcap API, I can do that too. I'm a bit undecided which solution is better.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,6 +14,12 @@ license = "MIT OR Apache-2.0"
 [dependencies]
 libc = "0.2.7"
 
+[features]
+# This feature enables access to the function Capture::savefile_append.
+# This is disabled by default, because it depends on a relatively recent
+# version of libpcap (1.7.2).
+pcap-savefile-append = []
+
 [lib]
 name = "pcap"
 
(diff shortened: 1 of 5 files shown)


maltek:
It looks like the `pcap_dump_open_append` function I used there was only added in  libpcap 1.7.2 early last year. So this won't work on older operating systems like the Ubuntu 12.04.5 LTS the continuous integration here uses.

Maybe this should indeed be a separate function, that can be optionally disabled?

fsck:
I also think that creating a separate function would be better. Coincidentally I have created that functionality (including a test) for an internal project a few month ago but have not yet created a pull request. Maybe we can work together on optionally disabling the function for Ubuntu 12.04.

You can look at my commit here: fsck/pcap@675c690

maltek:
Oh wow, that makes things easier. I've taken your code and put it behind a feature flag.

fsck:
Glad I could help. But the commit by aphistic is part of another pull request (#47) and I think you can safely leave it out of this one. I just needed it for my internal project.

ebfull:
Great! Sorry it took me so long to review this.

Excellent tests.

Issue #61, opened 2017-03-07
Allow linking to a custom libpcap/wpcap library


aldanor:
It would be nice if the downstream crates had an option to choose which libpcap to use -- e.g., in non-sudo environments. One way to do that would be to provide `build.rs` build script that would look for `PCAP_LIBDIR` environment variable and use that as `rustc-link-search` if it's set.

aldanor:
Closed by #62

PR #62, opened 2017-03-08
Support specifying libpcap location explicitly

aldanor:
* If `PCAP_LIBDIR` environment variable is set, add that to linker search path; this allows linking against a user-specified libpcap and not just the one at the default system-wide location.
* Minor: also relax libc requirement to just major/minor; this is more common than pinning the micro version and useful e.g. in workspaces.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ documentation = "https://ebfull.github.io/pcap/"
 license = "MIT OR Apache-2.0"
 
 [dependencies]
-libc = "0.2.7"
+libc = "0.2"
 
 [features]
 # This feature enables access to the function Capture::savefile_append.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,6 +10,7 @@ homepage = "https://github.com/ebfull/pcap"
 repository = "https://github.com/ebfull/pcap"
 documentation = "https://ebfull.github.io/pcap/"
 license = "MIT OR Apache-2.0"
+build = "build.rs"
 
 [dependencies]
 libc = "0.2"
--- a/README.md
+++ b/README.md
@@ -37,6 +37,10 @@ libpcap should be installed on Mac OS X by default.
 
 **Note:** A timeout of zero may cause ```pcap::Capture::next``` to hang and never return (because it waits for the timeout to expire before returning). This can be fixed by using a non-zero timeout (as the libpcap manual recommends) and calling ```pcap::Capture::next``` in a loop.
 
+## Specifying library folder
+
+If `PCAP_LIBDIR` environment variable is set when building the crate, it will be added to the linker search path - this allows linking against a specific `libpcap`.
+
 ## Optional Features
 
 To get access to the `Capture::savefile_append` function (which allows appending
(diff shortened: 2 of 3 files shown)


aldanor:
(Closes #61)

aldanor:
@ebfull Hi, wondering if you'd have time to review a few recent PRs?

I was actually planning to send many more :/

ebfull:
I've been super busy lately and I'm not sure that I have time to review PRs right now.

However, I'll be happy to give you some control over this repository so that PRs don't sit around. Is that okay?

Cargo.toml
```diff
@@ -10,9 +10,10 @@ homepage = "https://github.com/ebfull/pcap"
 repository = "https://github.com/ebfull/pcap"
 documentation = "https://ebfull.github.io/pcap/"
 license = "MIT OR Apache-2.0"
+build = "build.rs"
 
 [dependencies]
-libc = "0.2.7"
+libc = "0.2"
```
ebfull: Is this equivalent to `0.2.*`, I have forgotten the semver semantics.

README.md
```diff
@@ -37,6 +37,10 @@ libpcap should be installed on Mac OS X by default.
 
 **Note:** A timeout of zero may cause ```pcap::Capture::next``` to hang and never return (because it waits for the timeout to expire before returning). This can be fixed by using a non-zero timeout (as the libpcap manual recommends) and calling ```pcap::Capture::next``` in a loop.
 
+## Specifying library folder
+
```
ebfull: This deserves a "tips" or something section of the readme, where we can also put that note above this.

ebfull:
I'm +1 on this PR though.

Cargo.toml
aldanor: Yes, pretty much.

README.md
aldanor: Would there be anything else in this tips section?

aldanor: > I've been super busy lately and I'm not sure that I have time to review PRs right now.

> However, I'll be happy to give you some control over this repository so that PRs don't sit around. Is that okay?

@ebfull That'd be great, at least for the maintenance-like fixes and PRs. E.g., I was planning to integrate clippy and lint the code, maybe clean it up a little in general, that's pure technical task. I'd probably still open a PR and wait on some comments for anything discussion-worthy. Just a thought, would it make sense to move this out to an umbrella org? (e.g. pcap-rust/pcap, idk)

Also in the plans / half-implemented already - `pcap_*fopen*()` support for both reading/writing through raw fds on unix platforms (this is important e.g. if you want to read from / write to zipped pcaps; needs some careful testing though).

aldanor: (whoops, left the comment in a wrong thread)

ebfull: For now I'm happy keeping it where it is, but if it became a larger project it would be worth putting it in some kind of org. (With multiple repos.)

Adding you as a contributor.

aldanor:
I'll merge this PR since it's pretty straightforward (readme will address at a later time separately) and I'll probably open a few issues about larger design decisions -- would definitely appreciate your input if/when/whenever you have a minute. Thanks for adding me!

PR #63, opened 2017-03-09
Implement PartialEq and Eq for PacketHeader and Packet

aldanor:
(`libc::timeval` doesn't derive `PartialEq` so it has to be done manually)


ebfull:
This seems a bit fragile but okay.

PR #64, opened 2017-03-09
Add Capture::close() and Savefile::close() methods

aldanor:
This is a minor convenience but quite useful in some cases, so as not to introduce artificial scope blocks or do an explicit `drop()` just to close the handle.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -759,6 +759,11 @@ impl AsRawFd for Capture<Active> {
     }
 }
 
+impl<T: State + ?Sized> Capture<T> {
+    /// Closes the capture handle.
+    pub fn close(self) {}
+}
+
 impl<T: State + ?Sized> Drop for Capture<T> {
     fn drop(&mut self) {
         unsafe {
@@ -784,6 +789,9 @@ impl Savefile {
             raw::pcap_dump(*self.handle as *mut u8, transmute::<_, &raw::Struct_pcap_pkthdr>(packet.header), packet.data.as_ptr());
         }
     }
+
+    /// Closes the savefile.
+    pub fn close(self) {}
 }
 
 impl Drop for Savefile {


ebfull:
I would personally like to avoid code like this since it duplicates effort of the language (drop semantics) but isn't used for destruction/cleanup.

aldanor:
Thanks for the reply! It's more of a bikedshedding sort of question, so I won't regret too much if it doesn't get merged. However, I'll try to clarify my point :) This was inspired by actual code from the tests:

```rust
{
    // Scope for dead capture
    let dead_cap = pcap::Capture::dead(pcap::Linktype(1)).unwrap();
    let mut dead_save = dead_cap.savefile(&tmp_file).unwrap();
    for packet in &packets {
        dead_save.write(&packet);
    }
}

/* other code */
```

While the scope-based drop/close is nice so you don't have to write things manually, in this example it's pretty bad. The author of the code had to introduce a scope just to get something dropped; plus, if it wasn't for the comment, it's not immediately clear what *has* to be dropped here (`dead_save`).

An alternative solution would be

```rust
drop(dead_save);
```

which is something not everyone necessarily knows, plus explicit drops are not too nice, or do a  

```rust
let dead_cap = pcap::Capture::dead(pcap::Linktype(1)).unwrap();
let mut dead_save = dead_cap.savefile(&tmp_file).unwrap();
for packet in &packets {
    dead_save.write(&packet);
}
dead_save.close();  /* flush / close / drop */

/* other code */
```

In regards to

>  since it duplicates effort of the language (drop semantics) but isn't used for destruction/cleanup

an alternative way would be to move impl of `Drop::drop()` to `Capture::close()` and then just

```rust
impl Drop for Capture<...> {
    fn drop(self) { 
        self.close() 
    }
}
```

(note that `.close()` takes `self` and not a reference, so it has to be moved out)

Wojtek242:
This is an old PR, but in case it is considered it for inclusion I'll add my thoughts.

I think that rather than providing a `close` method which duplicates the `drop` semantics, a `flush` method could be added to resolve the underlying problem described in this issue. The `flush` method should act like its namesake in other libraries so dump all buffered content, but not to close.

Wojtek242:
Closed by #176

PR #66, opened 2017-03-10
Implement from_raw_fd(), from_raw_fd_with_precision(), savefile_raw_fd()

aldanor:
This adds support for both reading and writing to savefiles using raw descriptors on Unix platforms (+some tests refactoring and cleanup):

- `Capture<Offline>::from_raw_fd(fd: RawFd)`
- `Capture<Offline>::from_raw_fd_with_precision(fd: RawFd, precision: Precision)`
- `Capture<T>::savefile_raw_fd(fd: RawFd)`

The descriptors are opened via `fdopen()` and closed by libpcap. This piece of functionality is extremely helpful when reading/writing to e.g. gzipped pcaps -- the supplied tests emulate that use case.

A few minor open questions are: whether these new methods should be marked as `unsafe` and whether `FromRawFd` trait should be implemented.

An alternative would be to take in `*mut libc::FILE`, but that looks quite clunky in public API, and integer descriptor would be sufficient in most cases.

Note that there's large amounts of copy/paste in library code now, but I'd rather address that separately so as not to cram too much into a single PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,9 @@ build = "build.rs"
 [dependencies]
 libc = "0.2"
 
+[dev-dependencies]
+tempdir = "0.3"
+
 [features]
 # This feature enables access to the function Capture::savefile_append.
 # This is disabled by default, because it depends on a relatively recent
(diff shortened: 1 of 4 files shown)


aldanor:
Note: `FromRawFd` is not implemented for `Capture<Offline>` since it's supposed to return `Self` which would imply there would be no error handling.

PR #72, opened 2017-03-24
The big cleanup, part 1

aldanor:
This PR is a precursor to implementing automated version detection and splitting the crate into three.

- _Drastically_ reduced amount of copy/paste in the codebase.
- Enums like `Precision` etc are now repr(int), can use discriminants instead matching.
- Removed all non-libpcap bindings from `raw.rs`.
- Removed explicit panics (`get_datalink`, `list_datalinks`) -- we assume that `Capture<Active>` is activated and don't check for that in all the other places either, this seemed inconsistent.
- Removed most explicit unwraps and panics, folded them into the error type.
- Rename all `raw::Struct_foo` to `raw::foo`, remove extraneous typedefs in `raw`.
- Moved `PacketHeader` to `lib.rs`; `raw.rs` now only contains bindings.
- Always use `#[cfg(windows)]` to detect Windows platform.
- `Error::new()` now never fails/panics, even upon malformed UTF-8.
- Add support for running [`clippy`](https://github.com/Manishearth/rust-clippy), fix clippy issues.
- Run most code through [`rustfmt`](https://github.com/rust-lang-nursery/rustfmt).

@ebfull ^


ebfull:
Reviewing this tomorrow, wish me luck!

ebfull:
I did a high level pass and I'm very impressed. Good work! It can be merged if you're happy with it.

One bit of feedback: can we change Travis to run rustfmt and clippy against the codebase?

Don't block on me, but if I have some more time early this week I will try to do a more thorough review and look for any subtle coding mistakes.

aldanor:
Thanks -- yea, I'd merge this in the next few days so I can base the next big PR (version detection) off this branch. 

Re: Travis, clippy - I could definitely add (on nightly only). Rustfmt - I'd keep manual since it's not 100% ideal at this point and sometimes formats things in a weird fashion (e.g. I've reported an issue upstream while formatting this crate); for now it's easier to run `cargo fmt` and then inspect the diff manually. Most of the code here is now rustfmt-compliant.

aldanor:
@ebfull if there are no concerns re: this PR, I'd hit the green button :)

PR #74, opened 2017-06-08
Fix cstr_to_string

ericpts:
Previously the code was using CString::from_raw, however that was unsafe https://doc.rust-lang.org/beta/std/ffi/struct.CString.html#safety-1 and was assuming it had ownership of the string.

I converted it to https://doc.rust-lang.org/std/ffi/struct.CStr.html and added tests for linktype to make sure that it works.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -74,7 +74,7 @@ mod unique;
 /// An error received from pcap
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Error {
-    MalformedError(ffi::IntoStringError),
+    MalformedError(std::str::Utf8Error),
     InvalidString,
     PcapError(String),
     InvalidLinktype,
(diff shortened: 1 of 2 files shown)


ebfull:
Thanks!

I think travis is failing due to a clippy warning or some other reason.

src/lib.rs
```diff
@@ -744,7 +744,8 @@ fn cstr_to_string(ptr: *const libc::c_char) -> Result<Option<String>, Error> {
     let string = if ptr.is_null() {
         None
     } else {
-        Some(unsafe { CString::from_raw(ptr as _) }.into_string()?)
+        Some(unsafe {
+            CStr::from_ptr(ptr as _) .to_string_lossy().into_owned()})
```
ebfull: Is `to_string_lossy` a good idea? I think `to_string` with some error propagation would be better.

ebfull: Also, clippy is probably warning because you're doing .to_string_lossy() and .into_owned() inside of an unsafe block which isn't necessary.

aldanor: Agreed, it'd be more logical to have `CStr::from_ptr(ptr as _).to_str()?.into_owned()`.

ericpts:
Error::new() also SEGFAULTs because of this issue.

ericpts:
The CI clippy errors do not seem to be related to my changes:
```
error[E0432]: unresolved import `rustc::ty::MethodCall`

 --> /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/clippy_lints-0.0.134/src/mut_reference.rs:2:43

  |

2 | use rustc::ty::{TypeAndMut, TypeVariants, MethodCall, TyS};

  |                                           ^^^^^^^^^^ no `MethodCall` in `ty`
```

ebfull:
Thanks!

PR #76, opened 2017-07-04
Streaming

stephanbuys:
Based on the @Roxxik's referenced in #71.

PR #78, opened 2017-07-29
enable non blocking capture using selectable fd

lrbalt:
this pull enables  `set_nonblock` and `get_selectable_fd` so you can capture packets in an event loop

I have an example to use this using tokio. I'll put that in a separate pull request

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -660,6 +660,25 @@ impl<T: Activated + ?Sized> Capture<T> {
                 .map(|_| Stat::new(stats.ps_recv, stats.ps_drop, stats.ps_ifdrop))
         }
     }
+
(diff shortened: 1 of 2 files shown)


lrbalt:
sorry, I missed #76. I think this one differs from #76 only on error handling in get_selectable_fd

lrbalt:
failures are because of clippy not compiling on nightly

ebfull:
It looks like #76 contained everything from this PR as well, so this PR can be closed?

lrbalt:
Well, the error handling of get_selectable_fd differs from #76. And this enables the raw api methods ahead of #76 so you can use it now.

So I'd suggest to either adjust error handling on #76 or merge this one and rebate #76 on it.

lrbalt:
Closing because #76 was merged

PR #79, opened 2017-07-29
Non block example

lrbalt:
this is an example of using non blocking capture using the selectable fd of the capture device. It uses tokio to poll for new packets.

I'm not sure if the example is good enough for you. I could not get it to work to pass the captured packet into the reactor because of lifetime problems. I needed to create a owned copy to pass to the handler. Also this adds dependencies to tokio and mio for the example.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,9 @@ build = "build.rs"
 [dependencies]
 libc = "0.2"
 clippy = { version = "0.0.134", optional = true }
+futures = "*"
+mio = "*"
+tokio-core = "*"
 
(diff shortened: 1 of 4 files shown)


lrbalt:
Sorry, I completely missed PR #76.

Roxxik:
the other PR is still stuck in travis, maybe you wanna have a look at it and send a PR for the PR ;)

lrbalt:
failure is because clippy cannot be compiled on current nightly

lrbalt:
Closing because #76 was merged

PR #89, opened 2017-09-06
fix clippy feature

### LINKED ISSUE #82: Nightly build failure
I think this is due to the line saying `env: FLAGS="--features clippy,tokio"` when it should be `env: FLAGS="--features \"clippy tokio\""` (or something like that, modulo different escaping rules)

Trolldemorted:
The travis config tried to enable a feature named `clippy` that did not exist. Since we have a dependency called `clippy` we cannot define a feature named `clippy`, so i named it `clippy-lint`.

Fixes #82.


Wojtek242:
Resolved by #112

PR #91, opened 2017-10-22
Expose pcap_compile: compile pcap expressions to bpf

squeed:
This exposes pcap_compile, which is used in many libraries to compile pcap expressions to a bpf payload.

Also, remove explicit example listing; they're not needed anymore.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -77,6 +77,8 @@ use std::os::unix::io::{RawFd, AsRawFd};
 use self::Error::*;
 
 mod raw;
+pub use raw::bpf_insn;
+
 mod unique;
 #[cfg(feature = "tokio")]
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -754,6 +767,24 @@ impl Capture<Dead> {
             .map(|h| Capture::new(h))
             .ok_or(InsufficientMemory)
     }
+
+    /// Compiles a pcap filter expression to a bpf program, which is just
+    /// a list of bpf instructions
+    pub fn compile_filter(&mut self, program: &str) -> Result<Vec<bpf_insn>, Error> {
```
squeed: We could theoretically also implement this for other capture types, but I'm not sure of the utility.

PR #92, opened 2017-11-19
Use rust-bindgen to generate FFI bindings

tamird:


tamird:
@ebfull ping?

Trolldemorted:
Why would you want to reparse the headers with every build?

Won't this break builds on windows? Winpcap and npcap installations don't come with their headers.

tamird:
@Trolldemorted good point, and sorry for the delay. Updated this to include a script that regenerates the bindings, but the generated file is also included in the repo.

Wojtek242:
This PR is not compatible with the newest version of `pcap`. If you would like to have it included, please update and resubmit.

PR #94, opened 2018-02-12
Clarify timeout method doc comment

vignesh-sankaran:
Made a clarification to the timeout method on `Capture<Inactive>` since I spent a lot of time figuring out why the examples kept blocking for what appeared to be indefinitely on macOS.


Wojtek242:
This is an old PR and I'm not sure it clarifies anything. If you disagree, please update and reopen.

PR #95, opened 2018-07-18
Add documentation to undocumented items

bryanburgers:
Some of this documentation is manually written, some is taken from the
[pcap manpage][man] or the [pcap source][source].

These changes would make it possible to add `#![deny(missing_docs)]` to the module to always require a documented API, except that it doesn't document the Tokio features.

[man]: https://www.tcpdump.org/manpages/pcap.3pcap.html
[source]: https://github.com/the-tcpdump-group/libpcap/blob/master/pcap/pcap.h

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -84,17 +84,28 @@ pub mod tokio;
 /// An error received from pcap
 #[derive(Debug, PartialEq)]
 pub enum Error {
+    /// The underlying library returned invalid UTF-8
     MalformedError(std::str::Utf8Error),
+    /// The underlying library returned a null string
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -812,6 +894,7 @@ impl Drop for Savefile {
 }
 
 #[cfg(not(windows))]
+/// Open a raw file descriptor.
```
bryanburgers: Is this function intended to be public API? It looks like `Capture<Offline>::from_raw_fd` and `Capture<Activated>::savefile_raw_fd` would be the preferred public API.

PR #99, opened 2018-10-13
Port from tokio-core to tokio

kpcyrd:


sekineh:
#115 is the new attempt.

PR #106, opened 2019-10-18
pkg-config support

tsnoam:
`pkg-config` is the de-facto standard to retrieve compile and link information regarding system packages.

This PR adds basic support for pkg-config, while maintaining old behaviour for systems without pkg-config and particularly windows.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,9 @@ futures = { version = "0.1", optional = true }
 [dev-dependencies]
 tempdir = "0.3"
 
+[build-dependencies]
+pkg-config = "0.3.16"
+
 [features]
 # This feature enables access to the function Capture::savefile_append.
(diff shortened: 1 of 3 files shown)


Wojtek242:
The `build.rs` file has been updated quite a bit since this was submitted. We currently do version detection using an internal `libpcap` call. If you think `pkg-config` could benefit us could you please update the PR and resubmit.

tsnoam:
@Wojtek242 
First, I'm happy to see new maintainers had taken over. This is a good thing :)

At the moment, I don't have the time to play around with the `master` branch and thoroughly test everything. However, I did look at the git log and the relevant changes, and I suspect that linking with a static library (`libpcap.a` in linux, for example) would not work with the new approach.
Just wanted to bring this to your attention.

Issue #111, opened 2020-07-08
Add CHANGELOG.md file.


sekineh:
https://keepachangelog.com/en/0.3.0/

It's a good practice to have a `CHANGELOG.md` entry in each PRs.

Wojtek242:
Changelog added in #115

PR #115, opened 2020-08-25
Tokio update

### LINKED ISSUE #108: Updating version of tokio
Hi Team,

The version of tokio (tokio-core) being used is currently marked as archived for over 3 years. Can this project be migrated to use the newer crate?

Wojtek242:
Update `futures` and `tokio` to latest versions.

Closes #108 and #99 (which only updates to `tokio` 0.1 and doesn't update futures).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,9 @@ addons:
 jobs:
   include:
     - {os: linux, rust: stable,
-       env: FLAGS="--features tokio"}
+       env: FLAGS="--features capture-stream"}
     - {os: linux, rust: beta,
-       env: FLAGS="--features tokio"}
+       env: FLAGS="--features capture-stream"}
     - {os: osx, rust: stable,
        env: FLAGS="--features full" PCAP_LIBDIR=/usr/local/opt/libpcap/lib}
     - {os: osx, rust: beta,
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ build = "build.rs"
 libc = "0.2"
 clippy = { version = "0.0.*", optional = true }
 mio = { version = "0.6", optional = true }
-tokio = { version = "0.2", optional = true }
+tokio = { version = "0.2", features = ["io-driver"], optional = true }
 futures = { version = "0.3", optional = true }
 
 [dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ libc = "0.2"
 clippy = { version = "0.0.*", optional = true }
 mio = { version = "0.6", optional = true }
 tokio-core = { version = "0.1", optional = true }
-futures = { version = "0.1", optional = true }
+futures = { version = "0.3", optional = true }
 
 [dev-dependencies]
 tempdir = "0.3"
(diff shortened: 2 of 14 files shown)


src/lib.rs
```diff
@@ -47,12 +47,12 @@
 //! ```
 
 extern crate libc;
-#[cfg(feature = "tokio")]
+#[cfg(feature = "capture-stream")]
```
ebfull, line 50: These extern crates aren't necessary anymore now that the code is being switched to 2018 edition.

.travis.yml
```diff
@@ -7,9 +7,9 @@ addons:
 jobs:
   include:
     - {os: linux, rust: stable,
-       env: FLAGS="--features tokio"}
+       env: FLAGS="--features capture-stream"}
```
sekineh, line 10: This rename requires Changelog entry.

Cargo.toml
```diff
@@ -3,6 +3,7 @@
 name = "pcap"
 version = "0.7.0"
 authors = ["Sean Bowe <ewillbefull@gmail.com>"]
+edition = "2018"
```
sekineh, line 6: You might want to mention it in Changelog entry.

sekineh (changes requested):
Overall, this change renames feature name from `tokio` to `capture-stream` which is imcompatible to previous versions (`0.7.0`).
Is it inevitable? If so, please consider:

- Add [Changelog.md] entry (#111)
  - See https://keepachangelog.com/en/1.0.0/ for the file format.
  - Describe the reason for the feature name change and guide users to follow.
  - `0.7.0` section and older sections are OK to be left blank for now.

Wojtek242:
> Overall, this change renames feature name from `tokio` to `capture-stream` which is imcompatible to previous versions (`0.7.0`).
> Is it inevitable? If so, please consider:

Yes, I forgot to point it out. Cargo complains when a feature is called the same as a dependency. Previously, it wasn't a problem, because the dependency was on `tokio-core`, but now the dependency is on `tokio`.

I will add a changelog.

Cargo.toml
Wojtek242: Ah, I had this on my branch, but strictly speaking we don't need to update the version. I don't know what potential this has to be breaking though.

Trolldemorted:
How do you think about moving the async stuff into a separate crate (e.g. pcap-async) which depends on pcap/pcap-sys?

Wojtek242:
> How do you think about moving the async stuff into a separate crate (e.g. pcap-async) which depends on pcap/pcap-sys?

Potentially, but the first goal was to just bring everything up to date and resolve pending issues. Once that is done a slight organisational rejig wouldn't be off the table. If there's a good argument for it, want to open an issue?

Wojtek242:
Added a changelog as requested. Only question that remains is about the edition: a) shall I leave it at Rust 2018, b) shall I remove it , c) move it to its own issue, because it needs to be discussed

I would opt for a), because (from https://doc.rust-lang.org/edition-guide/editions/index.html):

> A Rust compiler will support all editions that existed prior to the compiler's release, and can link crates of any supported editions together. Edition changes only affect the way the compiler initially parses the code. Therefore, if you're using Rust 2015, and one of your dependencies uses Rust 2018, it all works just fine. The opposite situation works as well.

One reason to opt for b/c is that you would need a compiler that support Rust 2018 (>= 1.31 released in December 2018).

Thoughts?

Closes #111

sekineh:
I think it's Ok to include Edition 2018 in this PR. We need to specify minimum required rustc version in README until 0.8.0 release, though.

sekineh (approved):
Looks good to me!

Issue #120, opened 2020-08-28
cargo build --features tokio fails to build on windows


Trolldemorted:
```
C:\Users\Benni\repositories\pcap>cargo test  --features tokio
   Compiling pcap v0.7.0 (C:\Users\Benni\repositories\pcap)
error[E0432]: unresolved import `mio::unix`
 --> src\tokio.rs:3:10
  |
3 | use mio::unix::EventedFd;
  |          ^^^^ could not find `unix` in `mio`

error[E0412]: cannot find type `RawFd` in this scope
  --> src\tokio.rs:16:9
   |
16 |     fd: RawFd
   |         ^^^^^ not found in this scope

error[E0412]: cannot find type `RawFd` in this scope
  --> src\tokio.rs:49:37
   |
49 |     pub fn new(cap: Capture<T>, fd: RawFd, handle: &tokio_core::reactor::Handle, codec: C) -> Result<PacketStream<T, C>, Error> {
   |                                     ^^^^^ not found in this scope

error[E0432]: unresolved import `mio::unix`
 --> src\tokio.rs:3:10
  |
3 | use mio::unix::EventedFd;
  |          ^^^^ could not find `unix` in `mio`

error[E0412]: cannot find type `RawFd` in this scope
  --> src\tokio.rs:16:9
   |
16 |     fd: RawFd
   |         ^^^^^ not found in this scope

error[E0412]: cannot find type `RawFd` in this scope
  --> src\tokio.rs:49:37
   |
49 |     pub fn new(cap: Capture<T>, fd: RawFd, handle: &tokio_core::reactor::Handle, codec: C) -> Result<PacketStream<T, C>, Error> {
   |                                     ^^^^^ not found in this scope

error: aborting due to 3 previous errors

Some errors have detailed explanations: E0412, E0432.
For more information about an error, try `rustc --explain E0412`.
error: could not compile `pcap`.

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: aborting due to 3 previous errors

Some errors have detailed explanations: E0412, E0432.
For more information about an error, try `rustc --explain E0412`.
error: could not compile `pcap`.

To learn more, run the command again with --verbose.
```

sekineh:
I got the same error on my machine!

Oh in previous CI, error was ignored.... 
https://github.com/ebfull/pcap/runs/1039308893
[...]

Issue #119, opened 2020-08-28
test_raw_fd_api fails on ubuntu sometimes


Trolldemorted:
```
test test_raw_fd_api ... FAILED

failures:

---- test_raw_fd_api stdout ----
thread 'test_raw_fd_api' panicked at 'called `Result::unwrap()` on an `Err` value: PcapError("truncated dump file; tried to read 4 file header bytes, only got 0")', tests/lib.rs:238:34
stack backtrace:
   0: rust_begin_unwind
             at /rustc/397b390cc76ba1d98f80b2a24a371f708dcc9169/library/std/src/panicking.rs:475
   1: core::panicking::panic_fmt
             at /rustc/397b390cc76ba1d98f80b2a24a371f708dcc9169/library/core/src/panicking.rs:85
   2: core::option::expect_none_failed
             at /rustc/397b390cc76ba1d98f80b2a24a371f708dcc9169/library/core/src/option.rs:1221
   3: core::ops::function::FnOnce::call_once
   4: core::ops::function::FnOnce::call_once
             at /rustc/397b390cc76ba1d98f80b2a24a371f708dcc9169/library/core/src/ops/function.rs:227
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.


failures:
    test_raw_fd_api

test result: FAILED. 6 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out

error: test failed, to rerun pass '--test lib'
##[error]Process completed with exit code 101.
```
https://github.com/Trolldemorted/pcap/runs/1039588778?check_suite_focus=true

Wojtek242:
Are nightly builds something we should/want to support? Stable+beta yes. My impression was that the beta branch was meant for catching bugs before they hit stable. This crate was left on its own for 3 years and somehow it didn't break due to a compiler update. So is this nightly breakage really just the first one that happened since then?

sekineh:
If we move to edition 2018, specifying minimum supported rustc version (MSRV) is rather useful for users.

In that case, msrv (1.31 or something) + stable looks essential.

Trolldemorted:
>Are nightly builds something we should/want to support? Stable+beta yes. My impression was that the beta branch was meant for catching bugs before they hit stable.

[...]

PR #121, opened 2020-08-28
Improve github actions

Trolldemorted:
Some tweaks to your github actions:

- all operating systems and configurations in one matrix:
```
      matrix:
        os: [windows-latest, ubuntu-latest, macos-latest]
        profile: ['', --release]
        toolchain: [stable]
        features: ['']
```
- tests over all matrix variations (every os with every profile with every toolchain with every set of features)
- found two bugs already :tada: (#119, #120)
- daily scheduled build to detect problems with dependencies or the nightly compiler

I have deliberately not added the nightly toolchain and the features, because the build would fail due to the aforementioned bugs, but turning them on is just 1 loc change.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,49 +1,49 @@
-name: Rust
+name: Rust CI
 
 on:
   push:
     branches: [ master ]
   pull_request:
     branches: [ master ]
+  schedule:
+    - cron: "0 0 * * *"
 
 env:
+  RUST_BACKTRACE: 1
+  CARGO_TERM_VERBOSE: true
   CARGO_TERM_COLOR: always
-  LIB: 'c:\winpcap-dp\WpdPack\Lib\x64'
 
 jobs:
-  build:
-
-    runs-on: windows-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/rust.yml
```diff
@@ -1,49 +1,49 @@
...
-  build:
-
-    runs-on: windows-latest
+  cargo-test:
+    strategy:
+      matrix:
+        os: [windows-latest, ubuntu-latest, macos-latest]
+        profile: ['', --release]
+        toolchain: [stable]
+        features: ['']
```
sekineh, line 23: You disabled features testing (`tokio`, `full`).
If you enable `tokio` will CI fail? (Previously succeeded)

sekineh (approved):
Overall it looks like a very good infra!

Feature testing is crucial part of CI, please consider re-enable it again very soon.

.github/workflows/rust.yml
sekineh: Yes `--features tokio` fails. #120

PR #124, opened 2020-08-29
Fix all clippy and compiler warnings and add clippy to CI

### LINKED ISSUE #114: Resolve clippy issues
After #112, builds now run clippy, but it outputs warnings. See https://travis-ci.org/github/ebfull/pcap/jobs/720888528.

They should be fixed so that CI can be setup to fail on clippy warnings with `cargo clippy -- -D warnings`.

Wojtek242:
This PR fixes all clippy and compiler warnings. I've also added clippy to GitHub actions with the flag `-D warnings` which means the run will fail for any clippy warning.

Closes #114

### DIFF
--- a/tests/lib.rs
+++ b/tests/lib.rs
@@ -1,18 +1,25 @@
+#[cfg(not(windows))]
 use std::io;
 use std::ops::Add;
 use std::path::Path;
 use tempdir::TempDir;
 
-use pcap::{Active, Activated, Offline, Capture, Packet, PacketHeader, Linktype, Precision, Error};
+use pcap::{Active, Activated, Offline, Capture, Packet, PacketHeader, Linktype};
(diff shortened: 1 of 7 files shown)


tests/lib.rs
```diff
@@ -1,18 +1,25 @@
...
 #[cfg(windows)]
+#[allow(non_camel_case_types)]
 type time_t = libc::c_long;
 
 #[cfg(not(windows))]
+#[allow(non_camel_case_types)]
 type suseconds_t = libc::suseconds_t;
 #[cfg(windows)]
+#[allow(non_camel_case_types)]
 type suseconds_t = libc::c_long;
```
sekineh, line 23: Thanks for the fix. I've regretted the warning produced by these type aliases name.

sekineh (approved):

PR #129, opened 2020-08-31
Add features test for ubuntu and macos platforms / optimize num of jobs

### LINKED ISSUE #123: Remove Travis?
Should the `travis.yml` file be removed now that we use GitHub actions? #122 #121 #117

sekineh:
Closes #123 

I also reduced/optimized the number of jobs.


sekineh:
@Wojtek242 I think Github Actions now includes all the duty travis performed.
Please review.

Trolldemorted:
Is the amount of jobs a problem? Without beta&nightly testing on macos and windows you are blind to upcoming regressions.

README.md
```diff
@@ -48,6 +48,7 @@ If `PCAP_LIBDIR` environment variable is set when building the crate, it will be
 #### `capture-stream`
 
 Use the `capture-stream` feature to enable support for streamed packet captures.
+It's tested on ubuntu and macosx only.
```
Wojtek242, line 51: I think this should say. This feature is *supported* only on ubuntu and macosx. Otherwise, this gives the impression it should work on windows, but just isn't tested.

.github/workflows/rust.yml
```diff
@@ -18,9 +18,22 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
+        toolchain: [stable]
```
Wojtek242, line 21: I would at least keep beta

Wojtek242 (changes requested):

Wojtek242:
> Is the amount of jobs a problem? Without beta&nightly testing on macos and windows you are blind to upcoming regressions.

Good question. The amount of jobs was definitely an issue for Travis which only ran 5 jobs at a time and the MacOS jobs each took 5 minutes. However, GitHub actions seem to be faster so probably not as much of a problem.

The question of whether to test stable, beta, and nightly is different. I'm actually not a big fan of testing on nightly, because I don't really want to commit time to fixing nightly-only issues, especially since I have been always avoided the nightly toolchain. Testing on beta is valuable though as it does help spot upcoming regressions.

Trolldemorted:
>Good question. The amount of jobs was definitely an issue for Travis which only ran 5 jobs at a time and the MacOS jobs each took 5 minutes. However, GitHub actions seem to be faster so probably not as much of a problem.

Yeah travis is slow as hell, github actions allows 20 concurrent jobs. Macos and ubuntu are <1min per test, windows is sluggish because installing winpcap takes forever.

>The question of whether to test stable, beta, and nightly is different. I'm actually not a big fan of testing on nightly, because I don't really want to commit time to fixing nightly-only issues, especially since I have been always avoided the nightly toolchain. Testing on beta is valuable though as it does help spot upcoming regressions.

Consider yourself lucky, I have had a lot of dependencies that don't compile with stable/beta :|

How often do you expect failing nightly builds that require manual interaction?

README.md
sekineh: Did it as is. Thanks!

.github/workflows/rust.yml
sekineh: I personaly think 1.31 and stable are mandatory / must-have because they are _supported_.
beta and nightly are nice-to-have.

Unfortunately, 1.31.1 build failed, so we keep `stable` and `beta` for now.

sekineh:
Overall, I feel like _guilty_ when I put many jobs/minutes.
Github/MS is paying machine/electricity cost and it's not clear they are sustainably profitable.  If they see too much cost, they will have to re-think their strategy.
So I want to focus on the _supported_ platforms.

> How often do you expect failing nightly builds that require manual interaction?

I already met a nightly failure and needed to re-run the job manually. 
It's workable but adds an extra step.

sekineh:
@Wojtek242 
I think I have fixed it. Could you please merge (Squash and merge) this PR#?

PR #131, opened 2020-09-04
Add minimum supported rustc version.

### LINKED ISSUE #130: Pick working Minimum Supported Rustc Version
I tried 1.31.1 in the CI and unfortunately the build failed.
Result: https://github.com/ebfull/pcap/runs/1064490417

I feel backporting to the old compiler a bit counter-productive.

TODO:
- Pick the version we like to support (just picking the latest stable is OK, in my opinion)
- add GitHub Actions matrix for the version we chose.
- update the `README` which mentions `1.31`

Any opinions?

sekineh:
- Determined MSRV to be 1.40.0 as of this writing. Closes #130 .
- Add separate `cargo build` step.  It's useful to see where build failed. 
- Run clippy only when `stable` compiler.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
-        toolchain: ["1.39.0", stable]
+        toolchain: ["1.40.0", stable]
         profile: ['', --release]
         features: ['', '--features full']
         exclude:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
-        toolchain: ["1.40.0", stable, beta]
+        toolchain: ["1.43.1", stable, beta]
         profile: ['', --release]
         features: ['', '--features full']
         exclude:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
-        toolchain: [1.40, stable, beta]
+        toolchain: ["1.40", stable, beta]
         profile: ['', --release]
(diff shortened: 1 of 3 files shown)


Wojtek242 (changes requested):
Those jobs failed due to clippy warnings though (as I set clippy warnings to fail CI). The solution is to fix the warnings or to allow them. I think if it's easy enough to fix the warnings we should go for that as 1.44 is quite a recent version. If it's too much hassle, I'd change CI to not fail on clippy warnings and go with an older version of the compiler. 

Note, it's probably easier to do the bisection search if you disable clippy first and only fix the warnings once you find the actual oldest version of the compiler that still works.

sekineh:
> Those jobs failed due to clippy warnings though (as I set clippy warnings to fail CI). The solution is to fix the warnings or to allow them. I think if it's easy enough to fix the warnings we should go for that as 1.44 is quite a recent version. If it's too much hassle, I'd change CI to not fail on clippy warnings and go with an older version of the compiler.
> 

Hmm, do you suppose 1.44.1 is too new?

I don't think changing CI to not fail on clippy warning is a good idea in order to keep code healthy. In that case, we need to review clippy output manually. Because we have a lot of jobs, we won't have enough passion to see output by clicking details.

> Note, it's probably easier to do the bisection search if you disable clippy first and only fix the warnings once you find the actual oldest version of the compiler that still works.

Many projects run clippy on stable only.  Given clippy default changes over time and it's less important than build failures, I'd suggest to run clippy on stable only.
I think it's a reasonable trade off to get an earlier MSRV.

Wojtek242:
> Many projects run clippy on stable only. Given clippy default changes over time and it's less important than build failures, I'd suggest to run clippy on stable only.

Sounds reasonable to me since they pass on stable.

Wojtek242:
In that case just removing clippy from builds other than stable would mean we can support an older version of rustc.

sekineh:
My memo:
Dependency `bytes` 1.5.6 fails on 1.37.1. They don't have explicit MSRV claim.
https://github.com/ebfull/pcap/pull/131/checks?check_run_id=1074911356

sekineh:
I didn't know how to resolve review requests, I'm requesting new review from @Wojtek242 .

Issue #136, opened 2020-09-11
Build-time `pcap_lib_version` probe


sekineh:
It might be useful for linux / macos platforms. (automatically include `libpcap` version-dependent features)

Similar to:
https://github.com/Wojtek242/pcap/pull/3/files

sekineh:
I set `v0.8.0` milestone for now, assuming @Wojtek242 is eager to implement this.

sekineh:
Note:
In `npcap`, `pcap_lib_version()` probabry returns "1.9.x" or something, but they don't have `*fopen*` functions exported.
They have `*hopen*` functions instead.

See the link:
- https://gist.github.com/sekineh/31a6887ddca0ddf5b551e70436e2764d#gistcomment-3450129

Hmm, we still need `cfg(windows)` even after we implement version probe?

Wojtek242:
> Hmm, we still need `cfg(windows)` even after we implement version probe?

Sadly, that was always the case. For some reason the windows pcap library just does some things its own way. I don't think there is any way to avoid this.

Wojtek242:
> I set `v0.8.0` milestone for now, assuming @Wojtek242 is eager to implement this.

Yes, I will simply port the code from my fork. But let us decide first if we want `build.rs` or featurs in #118.

Wojtek242:
Resolved by #138

PR #138, opened 2020-09-19
Build-time libpcap version detection

### LINKED ISSUE #118: Evaluate usage of features
https://github.com/ebfull/pcap/blob/5eddf7f839d51374e23e9df62b4fdd31f04ed546/Cargo.toml#L25-L40

libpcap 1.7.2 was released in 2015 and 1.5 in 2013. Are there any distros still shipping those versions? Debian buster ships 1.8.1, CentOS 8 ships 1.9.

Wojtek242:
This pull requests replaces the previously existing per-API call features with build-time detection of libpcap version. There is also the option to specify the desired version using the `PCAP_VER` optional environment variable (the default is to use the version of the installed library).

Currently marked as a draft PR as I am not sure how to handle WinPcap. In #118 we discussed what the minimal supported libpcap version should be and 1.5.3 emerged as a possible candidate (supported on Centos 7 until 2024). Therefore, it would be ideal if I could also remove the `1_2_1` and `1_5_0` flags. However, I kept them around as the way WinPcap is currently handled is that it's treated as libpcap 1.0.0 with a few specific calls. I also don't know what npcap does at all for a version.

Questions for @sekineh:
- should we just keep handling WinPcap as a 1.0.0 library with some extra calls?
- can npcap be handled as a libpcap library version 1.9 (or whatever else it is)?
- or would it just be easiest to have a `wpcap` and `npcap` flags?

My proposal is:
- in `raw.rs` have a set of core functions that all of winpcap, npcap, and libpcap on linux share
- a config flag of the form `pcap_X_Y_Z` means that any of the three above that claim to support version X.Y.Z should work with this flag
- within a version we might have to split further between winpcap, npcap, and libpcap. For example in #136 it is pointed out that npcap has `hopen` instead of `fopen` so we would have:
```rust
#[cfg(pcap_1_5_0)]
extern "C" {
    #[cfg(not(windows))]
    { 
        pub fn pcap_fopen_offline_with_tstamp_precision(/* unix args */) -> *mut pcap_t;
    }
    
    #[cfg(npcap)]
    {
        pub fn pcap_hopen_offline_with_tstamp_precision(/* npcap args */) -> *mut pcap_t;
    }
 }
```

What do you think?

Closes #118, #136

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,22 +24,17 @@ futures = { version = "0.3", optional = true }
 tempdir = "0.3"
 tokio = { version = "0.2", features = ["rt-core"] }
 
+[build-dependencies]
+libloading = "0.6"
+regex = "1"
+
 [features]
-# This feature enables access to the function Capture::savefile_append.
-# This is disabled by default, because it depends on a relatively recent
-# version of libpcap (1.7.2).
-pcap-savefile-append = []
-
-# This feature enables access to the function Capture::from_raw_fd_with_precision.
-# This is disabled by default, because it requires libpcap >= 1.5.0.
-pcap-fopen-offline-precision = []
-
 # This feature enables access to the function Capture::stream.
 # This is disabled by default, because it depends on a tokio and mio
 capture-stream = ["mio", "tokio", "futures"]
 
 # A shortcut to enable all features.
-full = ["pcap-savefile-append", "pcap-fopen-offline-precision", "capture-stream"]
+full = ["capture-stream"]
 
 [lib]
 name = "pcap"
--- a/README.md
+++ b/README.md
@@ -43,6 +43,12 @@ libpcap should be installed on Mac OS X by default.
 
 If `PCAP_LIBDIR` environment variable is set when building the crate, it will be added to the linker search path - this allows linking against a specific `libpcap`.
 
+## Library Version
+
+The crate will automatically try to detect the installed `libpcap`/`wpcap` version by loading it during the build and calling `pcap_lib_version`. If for some reason this is not suitable, you can specify the desired library version by setting the environment variable `PCAP_VER` to the desired version. The version number is used to determine which library calls to include in the compilation.
+
+Th
+
 ## Optional Features
 
 #### `capture-stream`
@@ -53,29 +59,6 @@ This feature is supported only on ubuntu and macosx.
 ```toml
 [dependencies]
 pcap = { version = "0.7", features = ["capture-stream"] }
-```
-
-#### `pcap-savefile-append`
-
-To get access to the `Capture::savefile_append` function (which allows appending
-to an existing pcap file) you have to depend on the `pcap-savefile-append`
-feature flag. It requires at least libpcap version 1.7.2.
-
-```toml
-[dependencies]
-pcap = { version = "0.7", features = ["pcap-savefile-append"] }
-```
-
-#### `pcap-fopen-offline-precision`
-
-To enable `Capture::from_raw_fd_with_precision` constructor (which allows opening
-an offline capture from a raw file descriptor with a predefined timestamp precision)
-you have to add `pcap-fopen-offline-precision` feature flag. This requires libpcap
-version 1.5.0 or later.
-
-```toml
-[dependencies]
-pcap = { version = "0.7", features = ["pcap-fopen-offline-precision"] }
 ```
 
 ## License
--- a/README.md
+++ b/README.md
@@ -41,13 +41,11 @@ libpcap should be installed on Mac OS X by default.
 
 ## Library Location
 
-If `PCAP_LIBDIR` environment variable is set when building the crate, it will be added to the linker search path - this allows linking against a specific `libpcap`.
+If `LIBPCAP_LIBDIR` environment variable is set when building the crate, it will be added to the linker search path - this allows linking against a specific `libpcap`.
 
 ## Library Version
 
-The crate will automatically try to detect the installed `libpcap`/`wpcap` version by loading it during the build and calling `pcap_lib_version`. If for some reason this is not suitable, you can specify the desired library version by setting the environment variable `PCAP_VER` to the desired version. The version number is used to determine which library calls to include in the compilation.
-
-Th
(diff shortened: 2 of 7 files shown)


Wojtek242:
@sekineh will you have time to review? BTW, if you are too busy nowadays to work on `pcap` I understand. I'm happy to just wrap up the remaining bits of work for `0.8.0` myself and I'll leave the Windows parts as a TODO for later. Let me know : )

sekineh:
Oh I’ll look into it today!

Cargo.toml
```diff
@@ -24,22 +24,17 @@ futures = { version = "0.3", optional = true }
...
+regex = "1"
 
+[features]
 # This feature enables access to the function Capture::stream.
 # This is disabled by default, because it depends on a tokio and mio
 capture-stream = ["mio", "tokio", "futures"]
 
 # A shortcut to enable all features.
-full = ["pcap-savefile-append", "pcap-fopen-offline-precision", "capture-stream"]
+full = ["capture-stream"]
```
sekineh, line 37: We might want to add Changelog entries:

Deleted: 
* Feature flags "pcap-savefile-append", "pcap-fopen-offline-precision"

Added:
* Build-time libpcap version detection

Caveat: I'm not a native English speaker, please improve above descriptions!

src/lib.rs
```diff
@@ -435,6 +435,7 @@ impl Capture<Offline> {
 
     /// Opens an offline capture handle from a pcap dump file, given a path.
     /// Takes an additional precision argument specifying the time stamp precision desired.
+    #[cfg(pcap_1_5_0)]
```
sekineh, line 438: (Just a name nitpick)
We are preparing the crate `pcap-0.8.0` now, but will eventually see greater version numbers of our crate.
`pcap_1_5_0` might be confusing when we'll release `pcap-1.5.0` crate in future.

Suggestion: Use `libpcap` name when we refer to the C library we are wrapping.
* `pcap_1_5_0` -> `libpcap_1_5_0`
* `pcap_1_7_2` -> `libpcap_1_7_2`
* etc.

build.rs
```diff
@@ -1,7 +1,119 @@
...
+        Ok(Version::new(
+            major_str.parse::<usize>()?,
+            minor_str.parse::<usize>()?,
+            micro_str.parse::<usize>()?,
+        ))
+    }
+}
+
+fn get_pcap_lib_version() -> Result<Version, Box<dyn std::error::Error>> {
+    if let Ok(libver) = env::var("PCAP_VER") {
```
sekineh, line 40: Change `PCAP_VER` to `LIBPCAP_VER` to avoid mix up with the crate version?

README.md
```diff
@@ -43,6 +43,12 @@ libpcap should be installed on Mac OS X by default.
 
 If `PCAP_LIBDIR` environment variable is set when building the crate, it will be added to the linker search path - this allows linking against a specific `libpcap`.
 
+## Library Version
+
+The crate will automatically try to detect the installed `libpcap`/`wpcap` version by loading it during the build and calling `pcap_lib_version`. If for some reason this is not suitable, you can specify the desired library version by setting the environment variable `PCAP_VER` to the desired version. The version number is used to determine which library calls to include in the compilation.
```
sekineh, line 48: Change to `LIBPCAP_VER` to avoid mix up with the crate version?

README.md
```diff
@@ -43,6 +43,12 @@ libpcap should be installed on Mac OS X by default.
 
 If `PCAP_LIBDIR` environment variable is set when building the crate, it will be added to the linker search path - this allows linking against a specific `libpcap`.
```
sekineh, line 44: Just a note:
I'd never change this env var name to `LIBPCAP_LIBDIR`. I want to keep backward compatibility.

sekineh (commented):
Overall, I like how `raw` functions are grouped by `cfg` versions.

As a nitpick, I'd like to distinguish our crate and C library by different names.
* C library: want to call it `libpcap` wherever appropriate. Especially when specify versions.
* Rust crate: always call it `pcap`.

sekineh:
> Questions for @sekineh:
> 
> * should we just keep handling WinPcap as a 1.0.0 library with some extra calls?
> * can npcap be handled as a libpcap library version 1.9 (or whatever else it is)?
> * or would it just be easiest to have a `wpcap` and `npcap` flags?

I think it's OK to treat `WinPcap` as libpcap 1.0.0. PRs may come in future.
For `npcap` features, let's wait for PRs submitted by real `npcap` user.  

> * within a version we might have to split further between winpcap, npcap, and libpcap. For example in #136 it is pointed out that npcap has `hopen` instead of `fopen` so we would have:
> 
> ```rust
> #[cfg(pcap_1_5_0)]
> extern "C" {
>     #[cfg(not(windows))]
>     { 
>         pub fn pcap_fopen_offline_with_tstamp_precision(/* unix args */) -> *mut pcap_t;
>     }
>     
>     #[cfg(npcap)]
>     {
>         pub fn pcap_hopen_offline_with_tstamp_precision(/* npcap args */) -> *mut pcap_t;
>     }
>  }
> ```

As you wrote as above, we might be going to see `cfg(npcap)` flags detection in `build.rs` in future. If `npcap` returned higher version than WinPcap on windows platform, we could exploit that fact to detect which flavor of libraries is used.

src/lib.rs
Wojtek242: Good idea, done

Cargo.toml
Wojtek242: Done

build.rs
Wojtek242: Done. I also changed `PCAP_LIBDIR` to `LIBPCAP_LIBDIR` for consistency (added to changelog as well)

README.md
Wojtek242: Done

Wojtek242:
>  think it's OK to treat `WinPcap` as libpcap 1.0.0. PRs may come in future.
> For `npcap` features, let's wait for PRs submitted by real `npcap` user.

Good idea, let's wait for the users.

Wojtek242:
Merging as the review updates relate to naming only.

PR #139, opened 2020-10-17
Add rerun-if-env-changed to build.rs

Wojtek242:
If either of the `LIBPCAP` environment variables we use in `build.rs` changes, cargo will recompile


sekineh:
Lgtm, it seems to be a nice addition!

PR #140, opened 2020-10-17
Add support for immediate mode

### LINKED ISSUE #45: Why no pcap_set_immediate_mode / pcap_setmintocopy ?

### LINKED ISSUE #93: easylisten example hangs on macOS
I am currently running macOS 10.13, and the easylisten example hangs when attempting to capture packets and print them out. This is after I ran `sudo chmod o+r /dev/bpf*` to resolve an issue regarding BPF permissions.

Wojtek242:
The only complicated part of this change is how to support on WinPcap as WinPcap provides a separate API call `pcap_setmintocopy` which is used to set immediate mode. However, it does mean that the user can unset/set immediate mode by calling `min_to_copy` instead (possibly by accident).

Other options are:
* Not provide `min_to_copy` in `pcap`. However, since we must already provide `pcap_setmintocopy` in `raw.rs` then I see no reason not to.
* Make Windows user libraries use `min_to_copy` to set immediate mode and exclude the `immediate_mode` function from windows builds. I don't like this solution as this forces programmers to worry about windows vs linux/mac themselves.

Closes #45 
Should fix #93

### DIFF
--- a/examples/easylisten.rs
+++ b/examples/easylisten.rs
@@ -1,6 +1,14 @@
 fn main() {
     // get the default Device
-    let mut cap = pcap::Device::lookup().unwrap().open().unwrap();
+    let device = pcap::Device::lookup().unwrap();
+    println!("Using device {}", device.name);
+
+    // Setup Capture
+    let mut cap = pcap::Capture::from_device(device)
+        .unwrap()
+        .immediate_mode(true)
+        .open()
+        .unwrap();
 
     // get a packet and print its bytes
     println!("{:?}", cap.next());
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Add `Derive(Clone)` to `Device` struct (#100).
 - Build-time `libpcap` version detection.
+- Add support for immediate mode.
 
 ### Changed
 
--- a/examples/getdevices.rs
+++ b/examples/getdevices.rs
@@ -4,7 +4,11 @@ fn main() {
         println!("Found device! {:?}", device);
 
         // now you can create a Capture with this Device if you want.
-        let mut cap = device.open().unwrap();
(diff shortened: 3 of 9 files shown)


Wojtek242:
This is the last change from the fork I made a few months ago.

examples/easylisten.rs
```diff
@@ -1,6 +1,14 @@
 fn main() {
     // get the default Device
-    let mut cap = pcap::Device::lookup().unwrap().open().unwrap();
+    let device = pcap::Device::lookup().unwrap();
+    println!("Using device {}", device.name);
```
sekineh, line 4: I like this!

sekineh (approved):
Looks good, let’s add change log entry and merge this!

examples/easylisten.rs
Wojtek242: Yea, I found it harder to debug without this print statement!

PR #141, opened 2020-10-18
Join threads in tests

### LINKED ISSUE #126: Many tests don't care if threads panic
If the spawned threads are never joined, tests will not fail if they panic.

Wojtek242:
Only `test_raw_fd_api` used threads in the tests. The threads are now joined so that panics are propagated.

I would not squash these commits (just rebase) as strictly speaking the first commit fixes feature flags which is unrelated to the issue. Unless more changes are necessary for this issue it saves some hassle of creating multiple PRs.

Closes #126


sekineh (approved):
LGTM.

Issue #81, opened 2017-08-04
Capture::new should be unsafe{}


ebfull:
It should be unsafe, and also it should be renamed to `from_handle`.

Wojtek242:
Is unsafe as of #143

PR #143, opened 2020-10-18
Make all methods for constructing objects out of RawFd or raw pointers unsafe

### LINKED ISSUE #127: All from_raw_fd foo must be unsafe
As #119 and git blame have shown, even core devs cannot handle handle the unsafety 😛

Seriously, using them is unsafe and should be declared as such.

Wojtek242:
As per #127, constructing objects out of a RawFd is unsafe as the objects assume they are the sole owner of the file descriptor. This mimics the design of the [`FromRawFd`](https://doc.rust-lang.org/std/os/unix/io/trait.FromRawFd.html) trait. However, I decided not to implement that trait because
- we have more functions than just `from_raw_fd`
- we return a `Result` not `Self` - I figure it was preferable to keep the interface contract

If the trait is preferable in the future, we can think about it.

Closes #127.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@
 - `PCAP_LIBDIR` renamed to `LIBPCAP_LIBDIR` to distinguish the `pcap` crate
   from the `libpcap` library.
 - All methods that construct objects out of a `RawFd` are now unsafe.
+- All methods that take a raw pointer are now unsafe. Some of these functions
+  were renamed from `new` to `from_handle` to underline this.
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


sekineh (approved):
LGTM, both code and comments!

Wojtek242:
Hi @sekineh, thanks for the review! I have updated the code now, but I now also made all functions that take a raw pointer unsafe as well. After all these functions cannot guarantee that the pointers point to valid memory. I was inspired by #81 and I figured that I might as well make sure all unsafety is done correctly.

I'm new to this whole `unsafe` thing (I rarely wrote code where my own functions were `unsafe`) so I hope the changes are justified (hence asking for a new review).

sekineh (approved):
I've been thinking about this for a while.
The raw fd and raw pointers are two things that need to be `unsafe` in a sense supplier are supposed to pass a valid one.  It looks now clear after you described it in this PR#!

PR #145, opened 2020-10-27
Add First version of LinkName enum for Linktype

Stargateur:
I don't like much magic number, I follow the link of Linktype for value and name but I would appreciate some external verification of the code.

Unfortunately, if we want to add "Unknown" variant for the enum this would make it bigger (if we want to keep the unknown value), I don't known what could be the best but in anycase, this is useful. I wish rust could have a way to have a builting way to consider this kind of enum and make every other value fall under a Unknown magic variant that doesn't take place but keep the original value.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -225,24 +225,156 @@ impl<'a> Into<Device> for &'a str {
 /// This is a datalink link type.
 ///
 /// As an example, `Linktype(1)` is ethernet. A full list of linktypes is available
-/// [here](http://www.tcpdump.org/linktypes.html).
+/// [here](http://www.tcpdump.org/linktypes.html). The const bellow are not exhaustive.
+/// ```rust
+/// use pcap::Linktype;
+///
+/// let lt = Linktype(1);
+/// assert_eq!(Linktype::ETHERNET, lt);
+/// ```
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub struct Linktype(pub i32);
 
 impl Linktype {
     /// Gets the name of the link type, such as EN10MB
     pub fn get_name(&self) -> Result<String, Error> {
-        cstr_to_string(unsafe { raw::pcap_datalink_val_to_name(self.0) })
-            ?
-            .ok_or(InvalidLinktype)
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -229,6 +231,139 @@ impl<'a> Into<Device> for &'a str {
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
 pub struct Linktype(pub i32);
 
+#[derive(Debug, PartialEq, Eq, Clone, Copy, IntoPrimitive, TryFromPrimitive)]
+#[repr(i32)]
+pub enum LinkName {
```
sekineh, line 236: Although the definition of the `enum LinkName` is beautiful, I'm opposed to adding this because
- It represents the same thing as the existing `struct Linktype`. Providing two `pub` data types on the same concepts look confusing to me.
- It cannot represent `Linktype(290)` which might come from external / future pcap files as you have already pointed out.

My suggestion is to use the following associated `const`s instead:
```
impl Linktype {
    const NULL: Self = Linktype(0);
    const ETHERNET: Self = Linktype(1);
    ...
```

This pattern is also seen in [IPv6Addr::LOCALHOST](https://doc.rust-lang.org/std/net/struct.Ipv6Addr.html#associatedconstant.LOCALHOST) and other locations.

sekineh (commented):
@Wojtek242 , please feel free to add your comment.

sekineh:
We might want to add `impl Display for Linktype` which uses `get_name()` for easy inspection of data.

src/lib.rs
Stargateur: That not a bad idea

Stargateur: That not a bad idea

Stargateur:
> We might want to add `impl Display for Linktype` which uses `get_name()` for easy inspection of data.

That out of scope of the goal of my PR. But yeah why not, also get_name return a allocated String, that quite strange if like it's look pcap lib return a static data. First we should maybe have a get_name_ref that return a `&'static str`.

sekineh (commented):
LGTM, I think it deserves Changelog entry :-)

Wojtek242:
I manually merged this branch as another PR introduced a very minor conflict. This is on master now.

PR #146, opened 2020-10-29
Fix device lookups on windows

### LINKED ISSUE #110: `Device::lookup()` returns a bogus device on windows.
## Environment

Windows 10
WinPcap 4.1.3
rustup show > active toolchain
```
stable-x86_64-pc-windows-msvc (default)
rustc 1.44.0 (49cae5576 2020-06-01)
```
Cargo.toml
```
pcap = "0.7.0"
```

## Symptom

```
    let device = Device::lookup()?;
    println!("device = {:?}", device);
```
will print this:
```
device = Device { name: "\\", desc: None }
```

## Cause

The underlying `raw::pcap_lookupdev()` returns a wide string on Windows.
(The second byte is always `NUL`)

## Fix (PoC)

I quickly modified like below and it worked:
```
    /// Returns the default Device suitable for captures according to pcap_lookupdev,
    /// or an error from pcap.
    pub fn lookup() -> Result<Device, Error> {
        with_errbuf(|err| unsafe {
            let ptr = raw::pcap_lookupdev(err);
            wide_to_string(ptr as *const u16)
            // cstr_to_string(ptr)
                ?
                .map(|name| Device::new(name, None))
                .ok_or_else(|| Error::new(err))
        })
    }
(snip)

#[cfg(windows)]
fn wide_to_string(ptr: *const u16) -> Result<Option<String>, Error> {
    let string = if ptr.is_null() {
        None
    } else {
        let len = (0..256).position(|i| unsafe { *ptr.offset(i) } == 0).unwrap(); // FIXME: use Err instead of unwrap()
        let slice = unsafe {slice::from_raw_parts(ptr, len)};
        Some(OsString::from_wide(slice).to_string_lossy().into_owned())
    };
    Ok(string)
}
```

on my PC, I got this:
```
device = Device { name: "\\Device\\NPF_{A65B3E7F-4500-45AF-85EA-7ED788D547AD}", desc: None }
```

Trolldemorted:
Should fix #110

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,9 @@ mio = { version = "0.6", optional = true }
 tokio = { version = "0.2", features = ["io-driver"], optional = true }
 futures = { version = "0.3", optional = true }
 
+[target.'cfg(target_os = "windows")'.dependencies]
+widestring = "0.2"
+
 [dev-dependencies]
 tempdir = "0.3"
 tokio = { version = "0.2", features = ["rt-core"] }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -64,6 +64,9 @@ use std::os::unix::io::{RawFd, AsRawFd};
 
 use self::Error::*;
 
+#[cfg(target_os = "windows")]
+use widestring::WideCString;
+
 mod raw;
 mod unique;
 #[cfg(feature = "capture-stream")]
@@ -183,9 +186,18 @@ impl Device {
 
     /// Returns the default Device suitable for captures according to pcap_lookupdev,
     /// or an error from pcap.
+    #[cfg(not(target_os = "windows"))]
     pub fn lookup() -> Result<Device, Error> {
         with_errbuf(|err| unsafe {
-            cstr_to_string(raw::pcap_lookupdev(err))
+            cstr_to_string(raw::pcap_lookupdev(err))?
+                .map(|name| Device::new(name, None))
+                .ok_or_else(|| Error::new(err))
+        })
+    }
+    #[cfg(target_os = "windows")]
+    pub fn lookup() -> Result<Device, Error> {
+        with_errbuf(|err| unsafe {
+            wstr_to_string(raw::pcap_lookupdev(err))
                 ?
                 .map(|name| Device::new(name, None))
                 .ok_or_else(|| Error::new(err))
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -855,6 +867,17 @@ fn cstr_to_string(ptr: *const libc::c_char) -> Result<Option<String>, Error> {
     Ok(string)
 }
 
+#[cfg(target_os = "windows")]
+#[inline]
+fn wstr_to_string(ptr: *const libc::c_char) -> Result<Option<String>, Error> {
+    let string = if ptr.is_null() {
+        None
+    } else {
+        Some(unsafe { WideCString::from_ptr_str(ptr as _) }.to_string().unwrap())
```
sekineh, line 876: Thoughts on this `unwrap()`.

If we replaced `.to_string().unwrap()` with `.to_string()?` it won't compile because `Error` type are not compatible between caller and callee. We could modify our `Error` type, but I feel it like an overkill.

My suggestion is to simply replace `.to_string().unwrap()` with `.to_string_lossy()` [1] which always succeeds. What do you think about that?

[1] https://docs.rs/widestring/0.4.3/widestring/struct.UCString.html#method.to_string_lossy

Cargo.toml
```diff
@@ -20,6 +20,9 @@ mio = { version = "0.6", optional = true }
 tokio = { version = "0.2", features = ["io-driver"], optional = true }
 futures = { version = "0.3", optional = true }
 
+[target.'cfg(target_os = "windows")'.dependencies]
+widestring = "0.2"
```
sekineh, line 24: `0.4.2` looks like the latest.

sekineh (changes requested):
Thanks for the PR!
I think this PR deserve a Changelog entry, in the `Fixed` section.

src/lib.rs
Stargateur: pcap Error contains a MalformedError that represent Utf8 error why not use it ?

sekineh: It might be a good idea, but it's not a UTF8 string we are processing.
The best way here is probably: 
1) Add an error variant that wraps `FromUtf16Error`.
2) Return the above variant when error happens.

Wojtek242:
Any update on this? Once this is closed, I think it would be time to call it 0.8.0

sekineh:
It's OK to merge this and in that case I'll add another PR to tweak this.

PR #155, opened 2021-01-03
examples/getdevices.rs just lists devices

stappersg:
Commented out the capture code to prevent hang on no traffic devices.
So getdevices executable is closer to description of pcap project
feature list devices.


stappersg:
Seen the
> Some checks were not successful
> 24 cancelled, 3 failing, and 4 successful checks

My next merge request will honor `cargo fmt -- --check`

PR #160, opened 2021-01-24
Replace lookupdev with findalldevs

### LINKED ISSUE #158: `pcap_lookupdev` is obsolete
`Device::lookup` uses `pcap_lookupdev` which according to https://www.tcpdump.org/manpages/pcap_lookupdev.3pcap.html has been obsoleted.

Wojtek242:
Turns out there's already an example of using `findalldevs` in `pcap`. It was in the function below `list`. 

However, by removing `lookupdev` I realised that the case of no default device was not handled at all. I've currently handled this by returning a `Results<Option<Device>, Error>` as opposed to a `Result<Device, Error>`.

A possible alternative would be to translate the null option to an error, but that would be semantically different from what [libpcap does](https://www.tcpdump.org/manpages/pcap_findalldevs.3pcap.html) which does not consider the null case to be an error.

Currently a draft PR, because I want to have a think if there's a slightly prettier solution.

Closes #158

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
  - [doc](https://docs.rs/pcap/latest/pcap/) will now include all features
 
+### Changed
+
+- `Device::lookup` now returns `Result<Option<Device>, Error>` rather than `Result<Device, Error>`. `Ok(None)` means that the lookup succeeded, but no suitable devices were available. This is consistent with libpcap.
+
 ### Removed
 
 - `docs-rs` feature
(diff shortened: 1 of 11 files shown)


0b00001101:
I think the `Result<Options<Device>, Error>` return type makes more sense to me, especially since it matches what libpcap does. I'm not sure about a prettier way to do it, but the end result type works for me.

Wojtek242:
In this PR I remove `wstr_to_string` as the need for it came from a bug in the now obsolete `pcap_lookupdev`:
> In WinPcap and Npcap, this function may return a UTF-16 string rather than an ASCII or UTF-8 string. 
From https://npcap.com/guide/wpcap/pcap_lookupdev.html

This does not seem to apply to `pcap_findalldevs` on Windows: https://npcap.com/guide/wpcap/pcap_findalldevs.html

Therefore, this also closes #165.

PR #161, opened 2021-01-25
examples/getdevices.rs now just lists devices

stappersg:
Replaced capture code with a pointer to a "how to".  Reason for not
doing a capture on each device, is because it waits forever on devices
without network traffic.

This change also makes the example compatible with the
project feature of "list devices".


Wojtek242:
Awesome, thanks!

PR #162, opened 2021-02-14
Support for flushing pcap captures to disk

dglynos:
It's useful to be able to flush a pcap capture to disk. The patch provides support for this through Savefile.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1143,8 +1143,8 @@ impl Savefile {
 }
 
 impl Savefile {
-    pub fn flush<T>(&mut self, cap: &Capture<T>) -> Result<(), Error> 
-    where 
+    pub fn flush<T>(&mut self, cap: &Capture<T>) -> Result<(), Error>
+    where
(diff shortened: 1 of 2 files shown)


dglynos:
closing this as it's based on a master rather than a branch; will resubmit through a branch as it's cleaner

Trolldemorted:
Shouldn't this be done automatically in the Drop implementation?

Trolldemorted:
https://github.com/ENOFLAG/pcap/commit/ee44b9626dc4b392ab5e4750c8b49d539416a851 this is the commit we used

dglynos:
> Shouldn't this be done automatically in the Drop implementation?

Drop is great if you've got a capture that finished due to a user's action. However, if the capture was terminated for any other reason (an unexpected reboot, termination of a service etc.) and close() was not called on the file descriptor, then you're left with a file that may or may not contain any of the data captured. For this scenario, it's better to give users the chance to also flush when the application desires, not just at the end of a capture.

PR #163, opened 2021-02-16
generate optimized BPF program by default

thegwan:
tcpdump by default optimizes the generated bpf program, and I figured this crate should provide similar functionality.

An alternative could be to expose a parameter to the filter() and compile() methods to specify whether you want optimized programs.

### DIFF
--- a/examples/listenlocalhost.rs
+++ b/examples/listenlocalhost.rs
@@ -8,7 +8,7 @@ fn main() {
         .unwrap();
 
     // filter out all packets that don't have 127.0.0.1 as a source or destination.
-    cap.filter("host 127.0.0.1").unwrap();
+    cap.filter("host 127.0.0.1", true).unwrap();
 
     while let Ok(packet) = cap.next() {
         println!("got packet! {:?}", packet);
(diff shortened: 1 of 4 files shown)


Wojtek242:
I think it would be better to expose it as an option in my opinion since that's what `libpcap` does. `pcap` provides Rust abstractions on top of `libpcap`, but otherwise should just be the same API.

I'll hopefully fix the build failure tonight so if you update we can merge straight away.

thegwan:
I'll update the tests too at some point in the near future

thegwan:
@Wojtek242 any updates on the build failure fixes?

Wojtek242:
Yes, the builds are fixed on master. Your latest build fails due to `cargo fmt` though. You just need to reformat your code to satisfy the formatter.

Wojtek242:
This is a curious failure now. Possibly very unlucky and we'll see exactly the same on master on the next scheduled run. I will investigate it.

Wojtek242:
Yea, so it looks like you were just unlucky. Rust 1.51 was released that day which came with a new clippy lint. It did show up later in the scheduled runs.

I have now fixed it on master so can you please rebase? If it passes, we can finally merge it : ). Thanks a lot for your patience!

PR #168, opened 2021-03-30
Tell Tokio that pcap file descriptor is read-only

### LINKED ISSUE #90: tokio / streamlisten fails with IoError(InvalidInput)
Hi,

I've been trying to run the `streamlisten` example on OSX and it just fails with `IoError(InvalidInput)`.  The `easylisten` example runs through fine. I think this error comes up when the stream is executed, but it's hard to find out what might have caused this.

Happy to provide more info if you let me know what exactly you need :)

jvns:
Fixes #90, where the `streamlisten` example fails on Mac OS. 

This fix comes from [this comment](https://github.com/tokio-rs/tokio/issues/2413#issuecomment-616019772) on https://github.com/tokio-rs/tokio/issues/2413. It looks like the issue is that the pcap file descriptor is read-only, but PollEvented by default expects it to be read-write, so it raises an error.

I've tested that with this change, the `streamlisten` example works on Mac, and that without it I get an `InvalidInput` error when running that example.

### DIFF
--- a/src/stream.rs
+++ b/src/stream.rs
@@ -57,7 +57,7 @@ impl<T: Activated + ?Sized, C: PacketCodec> PacketStream<T, C> {
     pub fn new(cap: Capture<T>, fd: RawFd, codec: C) -> Result<PacketStream<T, C>, Error> {
         Ok(PacketStream {
             cap,
-            fd: tokio::io::PollEvented::new(SelectableFd { fd })?,
+            fd: tokio::io::PollEvented::new_with_ready(SelectableFd { fd }, mio::Ready::readable())?,
             codec,
         })
     }


Wojtek242:
Hi @jvns, thanks for the PR. I'd like to merge it, but it seems to be failing the pipeline on `cargo fmt`. Should be pretty straightforward to rectify.

Something for me to think of as a follow-up would be how to include tests for issues like the one you're fixing in the CI.

jvns:
Fixed!

DanielRussell:
Hi!  This change worked well for me, and also inspired me to update Tokio from 0.2 to 1.0.

Submitted as #170, which conflicts with this PR (and overwrites the changes here -- they touch the same lines of code).

jvns:
Awesome! Closing this then.

PR #170, opened 2021-04-07
Upgrade to use Tokio 1.0

DanielRussell:
* Change tokio::io::PollEvented to tokio::io::unix::AsyncFd

* Update streamlisten example

* Fix clippy needless_lifetimes lint

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -18,7 +18,7 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
-        toolchain: ["1.40.0", stable, beta]
+        toolchain: ["1.45.0", stable, beta]
         profile: ['', --release]
         features: ['', '--features full']
         exclude:
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,9 +25,9 @@ jobs:
           - os: windows-latest
             features: '--features full'
           # Tokio 1.0 (and this capture-stream) requires 1.45
-          - toolchain: "1.40.0"
-            features: '--features full'
           - toolchain: "1.45.0"
+            features: '--features full'
+          - toolchain: "1.40.0"
             features: ''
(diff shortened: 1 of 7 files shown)


DanielRussell:
I think CI is failing because [Tokio minimum supported version](https://github.com/tokio-rs/tokio#supported-rust-version) is 1.45 and CI uses 1.40.

Wojtek242:
Thanks for PR @DanielRussell! Apologies for keeping you waiting, but I didn't have much motivation for side projects in April. I'm back now though : )

So I'm not very keen on updating the minimum rustc version across the board to 1.45 since not that many distros support it (e.g. Debian Buster is on 1.41). At the same time I don't see much reason to support tokio 0.2 now that there is a 1.0 version.

Therefore, I'd rather require rustc 1.45 only for the optional `capture-stream` feature. This is actually quite straightforward to achieve using the GitHub workflows. Here's a commit you can cherry-pick: https://github.com/ebfull/pcap/pull/172/commits/ccfcde4b11d941b443c2de9ab8983b03157aa6fd (it's not merged into master).

Could you please incorporate that commit with the version split into your PR then? I've looked at the rest of the code and it looks fine to me.

DanielRussell:
Welcome back!

Great idea and solution @Wojtek242, thank you for laying out your thought process.

I think your patch mismatched the toolchain and features, so I updated.  I also removed `tokio` as a dev-dependency so the crate will build on 1.40.

Wojtek242:
They're not actually mismatched. Those are *excluded* workflows. What I wanted to do is to exclude `--features full` if the version is `1.40.0` (and vice versa for `1.45.0`). I've enabled the workflow to run and as you can see they fail.

Perhaps adding a comment to clarify this would help the next person working with these files.

Wojtek242:
Great, thanks for the PR!

Stargateur:
@DanielRussell https://github.com/ebfull/pcap/issues/187 ?

PR #197, opened 2022-01-10
Add auto traits as supertraits of State trait

not-my-profile:
Previously the following code:

    let mut cap = Capture::from_file("test.pcap").unwrap();
    let mut cap: Capture<dyn Activated> = cap.into();
    std::thread::spawn(move || {
        cap.next();
    });

failed with:

  `dyn Activated` cannot be sent between threads safely

because dyn Activated didn't implement Send.

Instead of only adding Send as a supertrait of Activated we can actually
add all the auto traits implemented by the Active, Dead, Inactive and
Offline enums as supertraits of the State trait, to cater to even more
scenarios.


Stargateur:
You should use generic, the is a little complex but you can see [here]https://docs.rs/pcap/latest/pcap/struct.Capture.html#impl-2) that [`next()`](https://docs.rs/pcap/latest/pcap/struct.Capture.html#method.next) is implemented for every [`Activated`](https://docs.rs/pcap/latest/pcap/trait.Activated.html). This mean you can do:

```rust
fn foo<State: Activated>(capture: Capture<State>) {
  capture.next();
}
```

This is way simpler than use dynamic dispatch.

not-my-profile:
Ah yeah of course ... thanks!

I updated this PR to just add an example about this to the docs.

Stargateur:
LGTM

Wojtek242:
Thanks @not-my-profile

PR #200, opened 2022-01-29
Tidy up the stream module

### LINKED ISSUE #192: Allow calling `sendpacket` after creating `PacketStream`
I would find it useful to be able to inject packets in between awaiting incoming packets. Guessing either (i) delegation of `sendpacket`; or (ii) adding a method `PacketStream::inner()` that gives a mutable reference to the underlying `Capture`; would do the trick.

axelf4:
* Remove unused direct dependency on mio.
* Inline the `Capture::next_noblock` method into stream module.
* Use newtype wrapper of `Capture` that implements `AsRawFd` with
  `pcap_get_selectable_fd()`, instead of new data type `SelectableFd` that
  carries around the fd;
* ...in order to use `AsyncFdReadyMutGuard::try_io` which abstracts away
  the details previously explicit in `Capture::next_noblock`.
* (Drop the `PacketCodec` requirement of being `Unpin`.)
* Add `PacketStream::inner_mut` to make it possible to inject packets
  in-between awaiting packets.

This is a breaking change since it removes `stream::SelectableFd` and
`stream::PacketStream::new` from the public API.

Closes #192

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,13 +63,9 @@
 
 use unique::Unique;
 
-#[cfg(feature = "capture-stream")]
-use core::task::Poll::Ready;
 use std::borrow::Borrow;
 use std::ffi::{self, CStr, CString};
 use std::fmt;
-#[cfg(feature = "capture-stream")]
-use std::io;
 use std::marker::PhantomData;
 use std::mem;
 use std::net::IpAddr;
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -1093,51 +1089,15 @@ impl<T: Activated + ?Sized> Capture<T> {
...
-
-                    Err(IoError(io::ErrorKind::WouldBlock))
-                }
-                Err(e) => Err(e),
-            }
-        }
-    }
-
-    #[cfg(feature = "capture-stream")]
+    #[cfg(all(unix, feature = "capture-stream"))]
```
Stargateur, line 1092: I somehow opposed to that, if a user require a feature disable it if not on unix will lead to uncepted error message such as "not find" instead of "can't compile on windows". This is misleading.

src/lib.rs
```diff
@@ -1230,17 +1190,28 @@ impl Capture<Dead> {
...
 #[cfg(not(windows))]
 impl AsRawFd for Capture<Active> {
+    /// Returns the file descriptor for a live capture.
     fn as_raw_fd(&self) -> RawFd {
-        unsafe {
-            let fd = raw::pcap_fileno(*self.handle);
+        let fd = unsafe { raw::pcap_fileno(*self.handle) };
+        assert!(fd != -1, "Unable to get file descriptor for live capture");
+        fd
+    }
```
Stargateur, line 1198: why force your style ? some people see unsafe as an unit the scope was hinting that untill we check fd value for erreur the code was unsafe (thus I like the change for assert)

Stargateur (commented):
I don't have knowledge of tokio/mio to know if the change is correct of not

src/lib.rs
axelf4: Returning an invalid fd cannot cause UB otherwise `as_raw_fd` would have to be marked as `unsafe`, so only the function call is unsafe. But I do not mind changing it back.

src/lib.rs
axelf4: Yup, I will revert this change. (Would be nice to be able to specify in the cargo manifest that some feature has some cfg requirement.)

src/lib.rs
Stargateur: Nevermind, it's impossible since the device should be active in this implementation of as_raw_fd, so assert is what what want

Wojtek242:
From a quick I like this PR. It does appear much cleaner. I would like to take a closer look at it first though.

However, for that can I ask you to do two things:
1) Since we don't have tests covering this feature, could you please add an example of the issue you were trying to solve in #192.
2) Can you make sure the CI passes - that way in case it takes me a bit too long to get around to reviewing I won't have to ping you for some minor fixes.

axelf4:
Hmm, I do not see how the recent Windows build failure would be caused by this PR.

Wojtek242:
@axelf4 thanks for the ping. The windows CI has been lately failing to download winpcap occasionally and needs to be given a kick to run again.

Looks like all tests pass. 

Thanks for the PR and the example! I'll have a look at the PR with your example some time soon.

Wojtek242:
Thanks for the PR. Code is nice and your example seems to work for me! I'll merge straight since I took my time and I don't want to block again, but if you have time, I'd appreciate if you could add a `CHANGELOG` entry for this PR.

PR #206, opened 2022-04-06
issue205

stappersg:
or an attempt to understand it

Issue #205, opened 2022-04-06
Feature capture-stream doc is not build on https://docs.rs/pcap/


Stargateur:
It would be better to build with all features in general doc.

```toml
[package.metadata.docs.rs]
all-features = true
```

BTW, we use `docs-rs` features for build.rs when we should probably use env or at least use standard `docsrs` feature https://docs.rs/about/builds#detecting-docsrs

stappersg:
@Stargateur is #206 what you mean?

If so: Make a merge request with better commit message as mine MR and which has you as committer.

Stargateur:
exactly thus I don't know if it would compile on docs.rs

stappersg:
On Wed, Apr 06, 2022 at 03:45:44AM -0700, Antoine wrote:
> exactly thus

Acknowledge


> I don't know if it would compile on docs.rs

Still transform your idea in merge request
(and close this issue)

PR #213, opened 2022-04-15
Compile pcap with all features for docs and remove docs-rs feature to…

Stargateur:
Closes https://github.com/rust-pcap/pcap/issues/205 and Closes https://github.com/rust-pcap/pcap/pull/206

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -20,13 +20,13 @@ jobs:
         os: [windows-latest, ubuntu-latest, macos-latest]
         toolchain: ["1.40.0", "1.49.0", stable, beta]
         profile: ['', --release]
-        features: ['', '--features full']
+        features: ['', '--all-features']
         exclude:
           - os: windows-latest
-            features: '--features full'
+            features: '--all-features'
           # Tokio 1.0 (and thus capture-stream) requires 1.49; do not run on 1.40
           - toolchain: "1.40.0"
-            features: '--features full'
+            features: '--all-features'
           - toolchain: "1.49.0"
             features: ''
         include:
@@ -34,7 +34,7 @@ jobs:
           - os: ubuntu-latest
             toolchain: nightly
             profile: ''
-            features: --features full 
+            features: --all-features
     runs-on: ${{ matrix.os }}
 
     steps:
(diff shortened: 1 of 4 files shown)


Stargateur:
honestly this lint sux https://rust-lang.github.io/rust-clippy/master/index.html#redundant_pattern_matching

.github/workflows/rust.yml
```diff
@@ -66,7 +66,7 @@ jobs:
       - name: cargo clippy
         if: matrix.toolchain == 'stable'
         run: cargo clippy ${{ matrix.profile }} ${{ matrix.features }} -- -D warnings
-      - name: cargo doc
+      - name: cargo doc --all-features
```
Wojtek242, line 69: This is just the name of the step, not the command that is to be run in the CI. The command to be run is below. What I recommend you doing is to keep the name as it was (as this same step is used when no features are enabled) and instead do the following:
1. Get rid of the `full` feature in `Cargo.toml`. Now that you have removed the `docs-rs` feature we can replace it with `--all-features` which is great!
2. In the CI file replace all instance of `--features full` to `--all-features` - this will also apply to the docs builds.

Also related to your change in `build.rs` I think it would be useful to prepend the command with `env DOCS_RS=1` so that we actually test the docs build in the environment `docs.rs` will use them.

Wojtek242 (changes requested):
Looks great! Thanks for the work!

I just have two small suggestions (see comment I left in `rust.yml`) to actually affect the CI and make sure we hit the `DOCS_RS` code path.

.github/workflows/rust.yml
Stargateur: facepalm.

> Get rid of the full feature in Cargo.toml. Now that you have removed the docs-rs feature we can replace it with --all-features which is great!

Unless I miss something there is no way to use `all-features` for dependency https://doc.rust-lang.org/cargo/reference/features.html#dependency-features. So there is still a reason to have a "full" features. I will follow your choice.

Wojtek242 (approved):

Wojtek242:
I've marked the changes as approved, but I'm hesitating, because the CI pipeline has not been triggered. Is there a way to check if the workflow file is okay?

Wojtek242:
Oops, I accidentally now pushed to main

Stargateur:
I don't understand why action didn't run. It's maybe because you request change. You can run one manually https://docs.github.com/en/github-ae@latest/actions/managing-workflow-runs/manually-running-a-workflow

Wojtek242:
I know why - I think it's because of renaming of the default branch

Wojtek242:
See #215

PR #232, opened 2022-04-18
Add a code coverage workflow

### LINKED ISSUE #211: Add code coverage to CI and publish on pages

Wojtek242:
Code coverage is pretty low, but the first step to improving it is to know that.

Closes #211

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -1,0 +1,49 @@
+name: Code Coverage
+
+on:
+  push:
+    branches: [ main ]
+
+env:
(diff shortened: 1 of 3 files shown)


Wojtek242:
The reason for using `cargo install` to install `grcov` is because the binary provided by `grcov` produces badges that show 0% as code coverage regardless of actual coverage. This is because the latest release is v0.8.7 from 11 Feb 2022. However, the fix to the calculation (https://github.com/mozilla/grcov/pull/782) is only included from 28 Feb. Despite there being tags for 0.8.8 and 0.8.9, no release is available for those versions.

PR #237, opened 2022-06-01
Sendqueue

qrnch-jan:
Implement support for sendqueue, which is a feature available in winpcap/npcap that allows Windows applications to send raw packets at a high sustained rate.


src/sendqueue.rs
```diff
@@ -0,0 +1,93 @@
...
+}
+
+impl SendQueue {
+    pub fn new(memsize: c_uint) -> Result<Self, Error> {
+        let squeue = unsafe { raw::pcap_sendqueue_alloc(memsize) };
+        let squeue = if let Some(squeue) = NonNull::new(squeue) {
+            squeue
+        } else {
+            return Err(Error::InsufficientMemory);
+        };
```
Stargateur, line 23: Could be `let squeue = NonNull::new(squeue).ok_or(Error::InsufficientMemory)?;`

src/sendqueue.rs
```diff
@@ -0,0 +1,93 @@
...
+    pub fn maxlen(&self) -> c_int {
+        unsafe { (*self.squeue.as_ptr()).maxlen }
+    }
+
+    pub fn len(&self) -> c_int {
+        unsafe { (*self.squeue.as_ptr()).len }
+    }
+
+    pub fn sync(&mut self, sync: bool) {
+        self.sync = if sync { 1 } else { 0 }
```
Stargateur, line 37: `bool` in Rust are guarantee to value 1 for true and 0 for false https://doc.rust-lang.org/std/primitive.bool.html

src/sendqueue.rs
```diff
@@ -0,0 +1,93 @@
...
+
+use libc::{c_int, c_uint};
+
+use crate::raw;
+use crate::Error;
+use crate::{Active, Capture};
+
+pub struct SendQueue {
+    squeue: NonNull<raw::pcap_send_queue>,
+    sync: c_int,
```
Stargateur, line 13: there is no reason to `sync` to not be `bool`

src/sendqueue.rs
```diff
@@ -0,0 +1,93 @@
...
+        let res = unsafe { raw::pcap_sendqueue_queue(self.squeue.as_ptr(), ph, buf.as_ptr()) };
+        if res == -1 {
+            return Err(Error::InsufficientMemory);
+        }
+
+        Ok(())
+    }
+
+    /// Transmit the contents of the queue.
+    pub fn transmit(&mut self, dev: &mut Capture<Active>) -> Result<(), Error> {
```
Stargateur, line 69: why don't you mimic pcap api and let user define `sync` bool here ?

src/sendqueue.rs
```diff
@@ -0,0 +1,93 @@
...
+        let res = unsafe {
+            raw::pcap_sendqueue_transmit(dev.handle.as_ptr(), self.squeue.as_ptr(), self.sync)
+        };
+
+        // ToDo: Fix unwrap()
+        if res < self.len().try_into().unwrap() {
```
Stargateur, line 75: the unwrap will be removed when you fix `pcap_send_queue` definition

src/sendqueue.rs
```diff
@@ -0,0 +1,93 @@
...
+        unsafe { (*self.squeue.as_ptr()).len }
+    }
+
+    pub fn sync(&mut self, sync: bool) {
+        self.sync = if sync { 1 } else { 0 }
+    }
+
+    /// Add a packet to the queue.
+    pub fn queue(&mut self, buf: &[u8]) -> Result<(), Error> {
+        let start = SystemTime::now();
```
Stargateur, line 42: do we want the user have more control over what we use as time ?

Stargateur (changes requested):

Stargateur:
I forget to add this require some tests

src/sendqueue.rs
qrnch-jan: - My original reasoning was that I would avoid even getting the time when a packet is added to the queue if `sync` is disabled.  This, however, did not happen -- but it could!
- From a usability point of view, if you know you need sync, you're going to know that you need to figure out how to set it.  If you don't know if you need need sync, it is extremely unlikely that you do, and that parameter will just add unnecessary confusion.  (I have conducted empirical experimentation on myself to come to this conclusion :)

I think this API design is better than the C one, but I can be convinced that it's better to mimic the original C API.

src/sendqueue.rs
qrnch-jan: The reason `sync` exists is basically to be able to send out packets with a high precision delay.  One might imagine that the use-case is primarily to perform replays/simulations.  One could further argue _if_ one would use this feature, it is likely that one would be loading packets from a capture, and would want to recreate the timings in the sendqueue from that capture buffer.  From that perspective you are completely right.

My use-case is not that though -- I'm interested in sending packets _**very**_ fast, and thus I do not use `sync` at all.  And I imagine that the replay case is far more niche than "send packets very fast".

Here's a suggestion:
- Add an `ts: Option<(i32, i32)>` for a timestamp to `queue()`
- if `sync` is set and:
  - `ts` is None, then get the system time is fetched as it is currently.
  - `ts` is Some, use the tuple's values as sec and usec instead.
- if `sync` is not set, then don't get time at all -- populate the time with zeroes.

The prior art:ish thing to do is to take in a pkthdr buffer to `queue`, since (as I noted above) that _if_ one uses this feature it is likely that one is populating the sendqueue from a capture buffer.  However, I _really_ want to set the lengths from the input slice, and thus the only interesting part of the header is the timestamp.

qrnch-jan:
The original sendqueue API is very .. spartan.

It doesn't have any reset at all, and the transmit function doesn't automatically reset.  Apparently if the transmit fails, the `len` is used to keep track of what packet failed.  So calling transmit again will retry from that point.  To reset one can manually set the `len` to 0.  (Which is what the crate's `reset()` does).

Also, I learned that the alloc/release functions don't actually allocate any magic buffer -- we could allocate any regular "Rust" buffer instead.  I used the sendqueue API's in case that changes in the future.  However, npcap's developer was quite clear about that it is safe to completely manage the buffer oneself.

Anywho; I'll work on the remaining points, and with regards to user-friendlieness -- I can make the transmit auto-reset on success.  Were there any other changes you had in mind?

src/sendqueue.rs
Stargateur: I think this time thing is a problem, first the code will stop working in something like ~15.5 years, secondly the user could want use monotonic time or whatever. https://github.com/rust-lang/libc/issues/1848 at the very least you should cast it to `time_t` and `suseconds_t` and we should probably let the user give us a `Duration`

src/sendqueue.rs
```diff
@@ -0,0 +1,94 @@
...
+        let s = since_the_epoch.as_secs();
+        let us = since_the_epoch.subsec_micros();
+
+        let pkthdr = raw::pcap_pkthdr {
+            ts: libc::timeval {
+                tv_sec: s as i32,
+                tv_usec: us as i32,
+            },
+            caplen: buf.len() as u32,
+            len: buf.len() as u32,
```
Stargateur, line 54: cast are wrong here just remove them

src/sendqueue.rs
qrnch-jan: The seconds in `timeval` is an `i32` in `libc` on Windows for some reason.  The wraparound is unlikely to ever be an issue if purely relative times are used.  Switched to use `Duration` as you suggested, which makes specifying relative times much more natural.

src/sendqueue.rs
qrnch-jan: I've added a try_into, since the `len()` is `usize` which is potentially longer than the `u32`.

Stargateur: ah yeah indeed much better

qrnch-jan:
> I forget to add this require some tests.

I'm having some trouble with this, because there's no obvious target to send packets to.

These updates are being developed for an application which has both a sender and receiver end, but this is not easily transformed to an isolated test case.

src/sendqueue.rs
```diff
@@ -0,0 +1,92 @@
...
+
+        Ok(Self(squeue))
+    }
+
+    pub fn maxlen(&self) -> c_uint {
+        unsafe { (*self.0.as_ptr()).maxlen }
+    }
+
+    pub fn len(&self) -> c_uint {
+        unsafe { (*self.0.as_ptr()).len }
```
Wojtek242, line 30: Aren't all your dereferincing calls like `(*self.0.as_ptr())` the same as `self.0.as_mut()`?

Wojtek242 (changes requested):
Can you please fix the clippy issues, and address the one comment I left?

At present it is difficult to add tests as there is no good testing framework for it, so please add an example instead which a user can run on their machine to verify it works.

Otherwise, since you've already addressed @Stargateur's comments it looks ready to merge.

qrnch-jan:
I've added an example which manually builds raw ethernet packets, queues them up and transmits them.  It uses a non-standard ethertype to avoid clashing with existing standard protocols.  One could argue that this could interfere with non-standard ethernet devices using ethertype 0x5555, but I figure that anyone looking to send raw ethernet packets is well aware of what this means.

The upside is that one can run a packet capture tool and filter on ethertype=0x5555 on the receiving system to easily see the transmitted packets, without getting noise from regular traffic.

I could make the ethertype configurable, but going by the other examples it looks like the examples are meant to be to-the-point (i.e. act as reference code).

Wojtek242 (approved):

PR #238, opened 2022-06-01
Wincrate

qrnch-jan:
Use the windows-sys crate rather than winapi.  Right now we're in the unfortunate situation where building Windows applications in Rust tends to eventually pull in multiple Windows API wrappers.  It _seems_ like the current trend is to convert projects to use windows{-sys} (since it's actively developed and supported by Microsoft).


Stargateur (approved):
why not thus the doc melt rustdoc https://docs.rs/windows-sys

But the last version of windows-sys doesn't fit the current pcap MSRV https://github.com/microsoft/windows-rs/blob/master/crates/libs/sys/Cargo.toml#L11

qrnch-jan:
Is the MSRV issue merely a documentation issue?  Should it be noted in `README.md` that Windows, without `capture-stream`, builds require `>= 1.46.0`?

qrnch-jan:
I updated the changelog and the readme, but I have never used github workflows before, so I'm fairly certain I borked something.

CHANGELOG.md
```diff
@@ -21,6 +22,7 @@
 ### Changed
 
 - `capture-stream` requires rustc version 1.49.0 due to dependency on `tokio`.
+>>>>>>> master
```
Wojtek242, line 25: Accidentally left over, please remove.

README.md
```diff
@@ -29,6 +29,8 @@ The feature `capture-stream` depends on `tokio`, but we only lock `tokio` versio
 
 ## Windows
 
+As of 0.9.3 Windows builds, without the `cature-stream` feature, requires a compiler version >= 1.46.0.
```
Wojtek242, line 32: Next release will `0.10.0`

Wojtek242 (changes requested):
Correct me if I'm wrong, but Windows doesn't really have something like Linux distributions where a particular package may be locked to a specific version in an LTS release for many years. In which case if `windows-sys` is really much better then it's okay having a different MSRV for Windows.

Anyway, some very minor changes and it can be merged.

CHANGELOG.md
sekineh: @Wojtek242 , did you mean to remove line 25 only? Or to remove both line 24 and 25?

Wojtek242 (approved):

Wojtek242:
Merged manually due to merge conflicts with sendqueue

PR #242, opened 2022-06-15
Make struct repr transparent so transmute is safe.

yotamofek:
A lot of the code assumes that `BpfInstruction` and `BpfProgram` have the same size and layout as their raw counterparts, but they have to have a `repr(transparent)` for that to be safe.

The [nomicon says](https://doc.rust-lang.org/stable/nomicon/transmutes.html):

> When transmuting between different compound types, you have to make sure they are laid out the same way! If layouts differ, the wrong fields are going to get filled with the wrong data, which will make you unhappy and can also be UB (see above).
>
> So how do you know if the layouts are the same? For repr(C) types and repr(transparent) types, layout is precisely defined. ...

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1369,7 +1369,9 @@ fn test_struct_size() {
     assert_eq!(size_of::<PacketHeader>(), size_of::<raw::pcap_pkthdr>());
 }
 
+#[repr(transparent)]
 pub struct BpfInstruction(raw::bpf_insn);
+#[repr(transparent)]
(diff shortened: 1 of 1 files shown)


yotamofek:
Here's one example:
https://github.com/rust-pcap/pcap/blob/bc8a3c0cbb8bcf4365fd959b9c79b0729fd79482/src/lib.rs#L1389
It's not a `std::mem::transmute`, rather a pointer cast, but the structs still need to be transparent for it to be safe :)

Stargateur (approved):
Agree,

PR #244, opened 2022-07-09
Add a into_state private method and remove 2 transmute

Stargateur:
This simply remove 2 unnecessary transmute.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -750,6 +750,14 @@ impl<T: State + ?Sized> Capture<T> {
         self
     }
 
+    fn into_state<S: State + ?Sized>(self) -> Capture<S> {
+        Capture {
+            nonblock: self.nonblock,
+            handle: self.handle,
+            _marker: PhantomData::<S>::default(),
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


Wojtek242:
Merged this manually due to merge conflicts

Wojtek242:
FYI, I had to revert this change as it was causing segmentation faults (see #254). However, if there is non-segmentation faulting way to implement this, it would be very welcome.

Stargateur:
I don't understand how that possible, I think this change reveal an UB somewhere else in the code.

Stargateur:
The problem is the Drop implementation:

```rust
impl<T: State + ?Sized> Drop for Capture<T> {
    fn drop(&mut self) {
        unsafe { raw::pcap_close(self.handle.as_ptr()) }
    }
}
```

Stargateur:
We should encapsulate `handle: NonNull<raw::pcap_t>,` into a type that have drop implement to avoid this trap it in future.

Wojtek242:
If you would like to submit a PR with it, it would be welcome

PR #250, opened 2022-08-14
Fix clippy lints

Wojtek242:


tests/lib.rs
```diff
@@ -316,13 +322,13 @@ fn test_compile() {
...
 
-    assert!(instructions.len() > 0);
+    assert!(!instructions.is_empty());
     assert!(program.filter(packet.data));
 
     let program = bpf_capture.compile("src host 8.8.8.8", false).unwrap();
     let instructions = program.get_instructions();
 
-    assert!(instructions.len() > 0);
+    assert!(!instructions.is_empty());
```
Stargateur, line 331: I don't like this lint, `!(!` is hard to read and far away for `is_empty()`.

Wojtek242: It's what clippy recommends so I'd prefer it for now. In general I do prefer using `is_empty` over `len > 0`. I get your point about `!(!` which is unfortunate. If you think it's a major impediment, please address this in a new PR in case there's a lengthy discussion about whether to or not disallow clippy lints. For now I will merge to make sure CI is passing again.

PR #252, opened 2022-08-15
Complete deviceinterface information

### LINKED ISSUE #13: Complete device/interface information
Stuff like address and so on is reported by `pcap_findalldevs`, we should store this within `Device`.

Wojtek242:
Closes #13 

It turns out that since the original issue was create 7 years ago almost everything has been added to the `Device` struct. The only thing remaining were the `flags` documented at: https://www.tcpdump.org/manpages/pcap_findalldevs.3pcap.html.

This field was slightly problematic as it consisted of two separate items really:
- interface flags, which were actual single-bit flags
- connection status, which was a 2-bit bitmask behind which were 4 enum-like values identifying the connection status

In the end I opted to create a wrapper struct that separates these two. My previous iteration without this wrapper is in commit 606bdc3.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -213,9 +213,8 @@ impl From<std::io::ErrorKind> for Error {
 }
 
 bitflags! {
-    /// Network device flags. For checking multiple flags at once, see the `bitflags` crate:
-    /// https://docs.rs/bitflags/latest/bitflags/.
-    pub struct DeviceFlags: u32 {
+    /// Network device flags.
+    pub struct IfFlags: u32 {
         /// Set if the device is a loopback interface
         const LOOPBACK = raw::PCAP_IF_LOOPBACK;
         /// Set if the device is up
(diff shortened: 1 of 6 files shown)


Wojtek242:
@Stargateur, would you mind doing a review?

src/lib.rs
```diff
@@ -203,23 +204,125 @@ impl From<std::io::ErrorKind> for Error {
     }
 }
 
+bitflags! {
+    /// Network device flags.
+    pub struct IfFlags: u32 {
```
Stargateur, line 209: I would not name it `IfFlags` but `DeviveFlags` or just `Flags`.

src/lib.rs
```diff
@@ -203,23 +204,125 @@ impl From<std::io::ErrorKind> for Error {
...
+            // DeviceFlags::CONNECTION_STATUS should be a 2-bit mask which means that the four
+            // values should cover all the possibilities.
+            _ => unreachable!(),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct DeviceFlags {
+    pub if_flags: IfFlags,
```
Stargateur, line 259: here I for sure want `flags:`

src/lib.rs
```diff
@@ -203,23 +204,125 @@ impl From<std::io::ErrorKind> for Error {
...
+    pub connection_status: ConnectionStatus,
```
Stargateur, line 260: `status` ?

Stargateur (approved):
look good to me, since it's just call at init that converting to enum is fine for me and more usable than put it in the bitfield. I didn't find any error.

src/lib.rs
Wojtek242: `status` is not clear enough and `connection_status` mirrors the original. I think this one is better left alone.

src/lib.rs
Wojtek242: That would potentially cause confusion with the `flags` field in `Device` leading to code like `device.flags.flags` which may be confusing. Happy to rename though, so open to other suggestions.

src/lib.rs
Wojtek242: Just `Flags` is not clear enough as there's also the `DeviceFlags` struct. `ConnectionFlags` does not reflect what it represents which is the interface itself which is why I've chosen `IfFlags`. However, I'm still open to other suggestions. Perhaps renaming `DeviceFlags` to `Flags` and the `IfFlags` to `DeviceFlags`?

src/lib.rs
Stargateur: we could choice to remove this intermediate struct and just put the flags and the connection status directly in Device.

Wojtek242: That was what I originally did. See the code here: https://github.com/rust-pcap/pcap/pull/252/files/606bdc33228ee8a386e739192837c14efb05a1d5. However, I do find that a bit uglier as pcap's `flags` contain two separate bits of information:
- the interface flags (LOOPBACK, UP, RUNNING, WIRELESS)
- connection status (which is not a flag but an enum of 4 values behind a bitmask)

I found the mix of these two a bit problematic and abstraction-wise I think separating them in wrapper struct is the `pcap` way of exposing nicer abstractions.

Wojtek242: The problem with the above approach is that `CONNECTION_STATUS` is not a flag, but a bitmask which made the unified struct internally inconsistent

Stargateur: bitflags can easy have flags that are not only one bit, if you wish you could remove the enum and put all in the bitflags struct.

PR #253, opened 2022-08-16
Build failure handle different libpcapso filename

Wojtek242:
The `build.rs` script is updated to no longer crash if it cannot find `libpcap` to query for its version. It will instead assume that the user is by default using the newest API version. For older API versions the user should consult the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,58 +19,79 @@ If you need anything feel free to post an issue or submit a pull request!
 * Write packets to savefiles
 * Inject packets into an interface
 
-See examples for usage.
+See [examples](examples) for usage.
 
 # Building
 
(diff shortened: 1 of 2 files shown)


Wojtek242:
So for example, if the user has a different libpcap.so file name the following might happen:
1. `libpcap` is at its newest version which means that the default assumed by `build.rs` is good enough.
2. `libpcap` is not at its newest which means the build will fail. The user should then consult the `pcap` README.
3. In the README the user finds that they can set `LIBPCAP_LIBDIR` and `LIBPCAP_VER`. The `LIBPCAP_LIBDIR` will fail since `build.rs` still assumes the `libpcap.so` name, but as a final fallback the user can set `LIBPCAP_VER`.

So in short, the original problem is solved. Either `libpcap` is at its newest version or the user has to set `LIBPCAP_VER`.

PR #260, opened 2022-08-16
Add Handle that own pointer and call C fn to free a pointer

Stargateur:
That an idea to avoid fall in trap of Drop impl like in https://github.com/rust-pcap/pcap/pull/244.

The purpose is to limit at maximum the risk of double free, with this `Handle` that own pointer it's very unlucky that a double free happen.

There is still one Drop impl of `BpfProgram` but this is a little more difficult since the free function ask for a struct not a pointer, this struct use a lot of unsafe, see https://github.com/rust-pcap/pcap/issues/261, if we are sure we can malloc to make the clone so we can use free yourself, meaning we could use `Handle::new(ptr, free);`.

With this PR https://github.com/rust-pcap/pcap/pull/244 should not introduce UB.

### DIFF
--- a/src/handle.rs
+++ b/src/handle.rs
@@ -1,0 +1,36 @@
+#![allow(dead_code)]
+
+use std::ptr::NonNull;
+
+/// MUST NOT IMPLEMENT COPY or CLONE
+/// This struct OWN the pointer
+/// and have a F implementation that will call F to free the pointer.
+#[derive(Debug)]
+pub struct Handle<T> {
+    handle: NonNull<T>,
+    drop: unsafe extern "C" fn(*mut T),
+}
+
+impl<T> Handle<T> {
+    pub fn new(handle: NonNull<T>, drop: unsafe extern "C" fn(*mut T)) -> Self {
+        Self { handle, drop }
+    }
+
+    pub fn as_ptr(&self) -> *mut T {
+        self.handle.as_ptr()
+    }
+
+    pub unsafe fn as_ref(&self) -> &T {
+        self.handle.as_ref()
+    }
+
+    pub unsafe fn as_mut(&mut self) -> &mut T {
+        self.handle.as_mut()
+    }
+}
+
+impl<T> Drop for Handle<T> {
+    fn drop(&mut self) {
+        unsafe { (self.drop)(self.as_ptr()) }
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -84,6 +84,8 @@ use windows_sys::Win32::Networking::WinSock::{AF_INET, AF_INET6, SOCKADDR_IN, SO
 #[cfg(target_os = "windows")]
 use windows_sys::Win32::Foundation::HANDLE;
 
+mod handle;
+use handle::Handle;
 mod raw;
 #[cfg(windows)]
 pub mod sendqueue;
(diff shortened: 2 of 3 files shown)


src/handle.rs
```diff
@@ -0,0 +1,36 @@
+#![allow(dead_code)]
```
Wojtek242, line 1: Why allow dead code? If `as_ref` and `as_mut` are not needed, they're not needed so I'd remove them. They can be added later when needed.

src/handle.rs
```diff
@@ -0,0 +1,36 @@
...
+
+use std::ptr::NonNull;
+
+/// MUST NOT IMPLEMENT COPY or CLONE
+/// This struct OWN the pointer
+/// and have a F implementation that will call F to free the pointer.
```
Wojtek242, line 7: Rewrite comment to be more formal and clearer. I suggest.
```
`Handle` MUST NOT implement `Copy` or `Clone` because it assumes it's the sole owner of the pointer and it will free it upon `Drop`ing.
```

src/lib.rs
```diff
@@ -1520,6 +1512,8 @@ fn test_struct_size() {
 
 #[repr(transparent)]
 pub struct BpfInstruction(raw::bpf_insn);
+
+// Be aware of the Drop impl
```
Wojtek242, line 1516: As per #261, please submit a PR to remove this, then rebase and we won't need this comment.

src/sendqueue.rs
```diff
@@ -19,7 +20,7 @@ impl SendQueue {
         let squeue = unsafe { raw::pcap_sendqueue_alloc(memsize) };
         let squeue = NonNull::new(squeue).ok_or(Error::InsufficientMemory)?;
 
-        Ok(Self(squeue))
+        Ok(Self(Handle::new(squeue, raw::pcap_sendqueue_destroy)))
```
Wojtek242, line 23: For consistency with `Capture` and `Savefile`, please implement a `From<NonNull<raw::pcap_send_queue>>` trait and use that.

Wojtek242 (changes requested):
Looks good. Left a few comments to improve consistency with the code base.

src/handle.rs
Wojtek242: Oh and it should be a normal comment (i.e. `//`) and not a documentation comment (i.e. `///`).

Wojtek242 (changes requested):
Oh and since we should start working towards more comprehensive testing, please add unit tests for `Handle`.

src/handle.rs
Stargateur: `sendqueue.rs` use them, I could change the code of sendqueue or add cfg feature gate just for them, I didn't like any of the solution.

src/lib.rs
Stargateur: unfortunately I can't use Handle for remove the `BpfProgram` `Drop` impl. I will see what I can do.

src/handle.rs
Wojtek242: Ah okay, leave a comment explaining this then.

Wojtek242:
Due to significant changes on `main` since this was proposed, I will close it. If you get around to addressing the comments and adapting to the current code base, I'll be happy to review it.

Issue #134, opened 2020-09-06
CI with npcap


Trolldemorted:
The windows tests are not run against npcap, because only [npcap OEM has a silent installer](https://github.com/ebfull/pcap/pull/117#issuecomment-681495849).

I [contacted the npcap project](https://github.com/nmap/npcap/issues/227), and they say we could work something out! I boldly assume you'll receive something you'll have to put into a github actions secret, so you might want to send the email yourselves, but if you are too busy I could do it too.

sekineh:
This info might be useful when both libraries are installed on the same machine:

## Note on WinPcap and npcap DLL search path (WinPcap comes first!)

https://nmap.org/npcap/guide/npcap-devguide.html#npcap-feature-native-dll

> Npcap installs its DLLs into C:\Windows\System32\Npcap\ instead of WinPcap's C:\Windows\System32\. Because of how Windows' DLL search path works, your application will use WinPcap first by default when Npcap and WinPcap coexist, as C:\Windows\System32\ is prior to C:\Windows\System32\Npcap\. So when Npcap and WinPcap coexist, an application that want to use Npcap instead of WinPcap must make C:\Windows\System32\Npcap\ precedent to the C:\Windows\System32\ in the DLL search path. Here are two ways to modify this search path to make your application load Npcap's DLLs first, based on how your application links Npcap/WinPcap's library (wpcap.dll).

sekineh:
In CI, just installing `npcap` only is enough.

Wojtek242:
Resolved by #273

PR #262, opened 2022-08-16
Use a more reliable command to DL winpcap on windows

Stargateur:
It's look `(New-Object System.Net.WebClient).DownloadFile` is a sort of async way to download file. Invoke-WebRequest is more "slow" according to what I found but that work perfectly for our need, it's basically curl. This should fix our problem.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,7 @@ jobs:
         if: matrix.os == 'windows-latest'
         run: |
           cinst -y winpcap --version 4.1.3.20161116
-          (New-Object System.Net.WebClient).DownloadFile("https://www.winpcap.org/install/bin/WpdPack_4_1_2.zip", "C:/WpdPack.zip")
+          Invoke-WebRequest -Uri "https://www.winpcap.org/install/bin/WpdPack_4_1_2.zip" -OutFile "C:/WpdPack.zip"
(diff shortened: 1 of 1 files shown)


Wojtek242:
I'm not sure this will fix the issue as the problem seems to be with a missing DLL, but windows is black magic to me so might as well try it.

Wojtek242:
Nope, still a problem: https://github.com/rust-pcap/pcap/runs/7892831644?check_suite_focus=true

Wojtek242:
I guess it's a problem with autohotkey - I remember when I was fixing the old pipeline when i took over that the Windows way of automatically installing exes was with another program that hits the right buttons on the GUI. Basically, if you want a CLI installer you need an msi file, and wpcap doesn't have one by default. Previously, I did find one on some random guy's website and it's probably still there. The correct solution would probably to move to npcap whose developers say they are willing to provide CLI installer access for open source projects.

Wojtek242:
AFAIK, you need to pay for the OEM version or [contact the npcap developers](https://github.com/rust-pcap/pcap/issues/134) as you would need `/S (Silent install, Npcap OEM only)`. If you would like to implement it, go ahead.

Stargateur:
> AFAIK, you need to pay for the OEM version or [contact the npcap developers](https://github.com/rust-pcap/pcap/issues/134) as you would need `/S (Silent install, Npcap OEM only)`. If you would like to implement it, go ahead.

pffff.... honestly npcap policy is...

PR #263, opened 2022-08-17
Fix version for bitflags - bump MSRV

rukai:
Using `bitflags = "~1.2"` will only allow bitflags versions matching `1.2.x`
Using `bitflags = "1.3.2"` will only allow versions above `1.3.2`.
And as mentioned [here](https://users.rust-lang.org/t/avoid-multiple-usage-of-the-same-crate/61865/3) cargo does not allow multiple semver compatible versions of the same crate, attempting to do so will fail to build.

This all means that if another dependency adds `bitflags = "1.3.2"` (a very common way to specify dependencies) then `cargo update`/`cargo run` will fail:
image

The easiest way to resolve this is to just bump the MSRV to support bitflags 1.3.

An alternative to the approach in this PR would be finding a way to remove the bitflags dependency entirely, feel free to close this PR and do that if you want.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,21 +18,12 @@ jobs:
     strategy:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
-        toolchain: ["1.41.0", "1.46.0", stable, beta]
+        toolchain: ["1.46.0", stable, beta]
         profile: ['', --release]
         features: ['', '--all-features']
         exclude:
           # capture-stream is not supported on Windows or compiler versions too old for Tokio.
           - os: windows-latest
-            features: '--all-features'
-          - os: windows-latest
-            toolchain: "1.41.0"
-          # 1.46.0 is only a special requirement on Windows
-          - os: ubuntu-latest
-            toolchain: "1.46.0"
-          - os: macos-latest
-            toolchain: "1.46.0"
-          - toolchain: "1.41.0"
             features: '--all-features'
         include:
           # nightly check is performed on ubuntu only.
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -18,22 +18,13 @@ jobs:
...
           - os: windows-latest
             features: '--all-features'
-          - os: windows-latest
-            toolchain: "1.41.0"
-          # 1.46.0 is only a special requirement on Windows
-          - os: ubuntu-latest
-            toolchain: "1.46.0"
-          - os: macos-latest
-            toolchain: "1.46.0"
-          - toolchain: "1.41.0"
```
Wojtek242, line 35: You should not remove the exclude of '--all-features' for MSRV as we don't support that. Only the base version is supported on MSRV. The other removals are fine.

Wojtek242 (changes requested):
> And as mentioned [here](https://users.rust-lang.org/t/avoid-multiple-usage-of-the-same-crate/61865/3) cargo does not allow multiple semver compatible versions of the same crate, attempting to do so will fail to build.

Ay, I was not aware of this. I thought that cargo will always just install multiple versions.

This PR is fine as the bump is only up to 1.46 which is still supported by all officially Debian versions. There's just one fix you need to make to the ci file.

Wojtek242:
Oh, and it would be great if you could make a note of the MSRV bump in the changeling.

rukai:
> Ay, I was not aware of this. I thought that cargo will always just install multiple versions.

Yeah its a bit of a hidden footgun :/

Thanks for the feedback, glad we can go ahead with this, everything should be addressed now.

Wojtek242 (approved):

Wojtek242:
Thanks for the PR! FYI, I have just published 0.10.1 which includes your fix.

Issue #264, opened 2022-08-17
SendQueue shouldn't be exposing c_uint


Wojtek242:
This crate is meant to hide the C behind a Rust API

PR #268, opened 2022-08-23
Don't use c_uint in public interface.

qrnch-jan:
Fixes https://github.com/rust-pcap/pcap/issues/264


Wojtek242 (changes requested):
Looks great! Thanks! You missed just one `c_uint` and after that I can merge.

Wojtek242 (approved):

PR #273, opened 2022-08-29
Npcap CI

### LINKED ISSUE #196: Paths for Upgrading Winpcap to Npcap for Microsoft Windows
In an effort to consolidate our efforts to update the crate to use Npcap over winpcap, I made this ticket so we can centralize any discussion. @Wojtek242 has put forth [two ideas](https://github.com/ebfull/pcap/pull/190#issuecomment-934777143) for upgrading the use of winpcap to Npcap:

1. Update the documentation to support both winpcap and Npcap options and note that the CI pipeline still relies on winpcap
2. Fully replace winpcap with npcap and update the CI pipeline to use Npcap

I was considering a third option that combines both of the above:

3. Update the documentation and CI to use Npcap with winpcap compatibility mode, until we fully remove the old winpcap references
    - Silently install Npcap with the `/winpcap_mode` switch
    - This would also let us get the new updates and improvements that Npcap contributed towards the winpcap libraries

This allows us to keep the same code, but only update the documentation and CI pipeline until we fully remove the dependence on winpcap. The only issue is that we need to get the paid version of Npcap (Npcap OEM). Fortunately, another contributor [asked the nmap folks](https://github.com/nmap/npcap/issues/227) and they mentioned that we just need to email them to get it sorted out.

From my limited testing, installing Npcap with winpcap compatibility worked without any issues. The tests completed successfully and I was able to get a list of devices.

Thoughts? I can reach back out to nmap to get this started if we want to go that route or we can just wait until we fully remove winpcap in one go.

Wojtek242:
Closes #196 #134

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,16 +44,21 @@ jobs:
       - name: Install libpcap (macos)
         if: matrix.os == 'macos-latest'
         run: brew install libpcap
-      - name: Install WinPcap and WinPcap 4.1.2 Developer's Pack (windows)
+      - name: Install npcap sdk (windows)
         if: matrix.os == 'windows-latest'
+        run: |
+          Invoke-WebRequest -Uri "https://npcap.com/dist/npcap-sdk-1.13.zip" -OutFile "C:/npcap-sdk.zip"
+          Expand-Archive -LiteralPath C:/npcap-sdk.zip -DestinationPath C:/npcap-sdk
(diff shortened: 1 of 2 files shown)


Wojtek242:
Disclaimer: 
> Secrets are not passed to workflows that are triggered by a pull request from a fork.
and since we need secrets to silently install npcap with the oem version we cannot run tests using npcap in pull requests from public forkrs. Therefore, windows test builds are now restricted to after PRs have been accepted. Since we still build the windows binaries and Linux and MAC OS X still run the tests that's acceptable for a bindings library.

PR #277, opened 2022-09-29
Pick up libpcap from pkg-config if available.

### LINKED ISSUE #276: pkg-config support
It would very nice if `pkg-config` crate could be used to probe for location and version of `libpcap`. This can work a lot better than specifying LIBPCAP_LIBDIR and LIBPCAP_VERSION manually in a lot of scenarioes.

I can send a PR if desired, even with a feature gate.

Fuuzetsu:
To keep existing behaviour, it will look for `LIBPCAP_LIBDIR` first and only then go the pkg-config way and then go for the hail mary as it used to. This will be less likely to break any existing setup and is similar to `pq-sys` for example.

Fixes #276

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -172,14 +172,24 @@ fn main() {
     let version = if let Ok(libdir) = env::var("LIBPCAP_LIBDIR") {
         println!("cargo:rustc-link-search=native={}", libdir);
         get_libpcap_version(Some(PathBuf::from(&libdir))).unwrap()
-    } else if let Ok(library) = pkg_config::Config::new()
-        .atleast_version(&Version::LOWEST_SUPPORTED.to_string())
-        .probe("libpcap")
-    {
+    } else if let Ok(library) = from_pkg_config() {
         Version::parse(&library.version).unwrap()
     } else {
         get_libpcap_version(None).unwrap()
     };
 
     emit_cfg_flags(version);
+}
+
+fn from_pkg_config() -> Result<pkg_config::Library, pkg_config::Error> {
+    let mut config = pkg_config::Config::new();
+    // If the user has went out of their way to specify LIBPCAP_VER (even though
+    // LIBCAP_LIBDIR wasn't set), respect it. Otherwise fall back to any version
+    // as long as it's supported.
+    if let Ok(v) = env::var("LIBPCAP_VER") {
+        config.exactly_version(&v);
+    } else {
+        config.atleast_version(&Version::LOWEST_SUPPORTED.to_string());
+    };
+    config.probe("libpcap")
 }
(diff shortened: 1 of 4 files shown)


build.rs
```diff
@@ -148,12 +165,21 @@ fn main() {
...
+    let version = if let Ok(libdir) = env::var("LIBPCAP_LIBDIR") {
         println!("cargo:rustc-link-search=native={}", libdir);
-        libdirpath = Some(PathBuf::from(&libdir));
-    }
+        get_libpcap_version(Some(PathBuf::from(&libdir))).unwrap()
+    } else if let Ok(library) = pkg_config::Config::new()
+        .atleast_version(&Version::LOWEST_SUPPORTED.to_string())
+        .probe("libpcap")
+    {
+        Version::parse(&library.version).unwrap()
```
Stargateur, line 179: Shouldn't this also have something like:

```rust
for link_path in library.link_paths {
  println!("cargo:rustc-link-search=native={}", link_path);
}
```

Wojtek242: Doesn't look like it:

> After running pkg-config all appropriate Cargo metadata will be printed on stdout if the search was successful.

From https://docs.rs/pkg-config/latest/pkg_config/

Wojtek242:
Great! Thanks a lot! 

Could I ask you to just wrap this up with an update to the CHANGELOG.md file and to the README sections "Library location" and "Library version" to explain the use of pkg config. It might be worth pointing out that this can be used to configure static linking which some people are quite interested in (https://docs.rs/pkg-config/latest/pkg_config/).

Just a few sentences and don't worry about word smithing. I can modify/polish it later myself. There's also no rush, so whenever you have time.

build.rs
```diff
@@ -148,12 +165,21 @@ fn main() {
...
+    // existing build scripts running. If it's not set, try pkg-config. If
+    // that's not set, try last ditch effort to build even though library wasn't
+    // explicitly given.
+    let version = if let Ok(libdir) = env::var("LIBPCAP_LIBDIR") {
         println!("cargo:rustc-link-search=native={}", libdir);
-        libdirpath = Some(PathBuf::from(&libdir));
-    }
+        get_libpcap_version(Some(PathBuf::from(&libdir))).unwrap()
+    } else if let Ok(library) = pkg_config::Config::new()
+        .atleast_version(&Version::LOWEST_SUPPORTED.to_string())
```
Fuuzetsu, line 176: I wonder if we should check for LIBPCAP_VER here and only default to LOWEST_SUPPORTED if it's unset. This would allow "picking" a version

Fuuzetsu: I decided to do this. We can revert it if it's dumb.

Fuuzetsu:
Added some words to README and CHANGELOG + added a change to honour `LIBPCAP_VER` even in pkg-config scenario.

PR #283, opened 2022-12-09
feat: fix and test examples in docs

lockness-Ko:
Added an extra unwrap to unwrap `Option<Device>` in both examples.

Just a minor addition to help newbies to rust :smiley:.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,11 +8,10 @@
 //! use `.open()` on a `Device`. You can obtain the "default" device using
 //! `Device::lookup()`, or you can obtain the device(s) you need via `Device::list()`.
 //!
-//! ```ignore
+//! ```
 //! use pcap::Device;
-//!
 //! fn main() {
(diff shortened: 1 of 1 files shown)


Wojtek242:
Unfortunately, it looks like due to permission errors this test cannot be added.

I have two suggestions:
1. Either you can figure out how to can around the permission issues, OR
2. You re-add the `ignore` label, but leave the other changes.

lockness-Ko:
No worries! I think I'll re-add the ignore label and add a footnote to mention that to examples actually _should_ work (in normal circumstances).

robs-zeynet:
Note that the permissions issue is the same as in #285 which I've tried to dodge by using a feature flag instead of the 'ignore' option.  Feedback welcome, but IMHO the feature flag is a cleaner way to say "this test should work".

Wojtek242:
Resolved by #296

PR #285, opened 2022-12-29
Initial Linux-only virtual interface testing

robs-zeynet:
This is an end-to-end test of rust pcap and the underlying libpcap and OS. Currently it only works on Linux (and maybe MacOS), must be run with root (at least CAP_NET_ADMIN) permissions.

The test creates a 'TAP' interface
(https://en.wikipedia.org/wiki/TUN/TAP) which is a virtual NIC on one side and a file descriptor on the otherside (to fascilitate testing).

I've hidden it behind a 'tap-tests' cargo feature try to prevent people from accidentally compile it (which requires the 'tun-tap' and 'etherparse' additional dev dependencies) or running it without meeting the above requirements.

Current test just verifies that packets can be sent and received, e.g., do Capture<Active>::next_packet() and Capture<Active>::sendpacket() do the right things.

A fair bit of work is done to avoid race conditions which can be horrible for these types of tests.

	[robs@laptop-o7rs71ij pcap]$ cargo test -F tap-tests --no-run --test
	tap_tests | tail
	    Finished test [unoptimized + debuginfo] target(s) in 0.05s
	  Executable tests/tap_tests.rs
	(target/debug/deps/tap_tests-232532a0859d5255)
	[robs@laptop-o7rs71ij pcap]$ sudo
	./target/debug/deps/tap_tests-232532a0859d5255

	running 1 test
	test tests::conntrack_tap_basic ... ok

	test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered
	out; finished in 0.54s

### DIFF
--- a/tests/tap_tests.rs
+++ b/tests/tap_tests.rs
@@ -1,0 +1,149 @@
+/***
+* These tests need to be run as root and currently only work on Linux (and maybe MacOs?)
+*
+* To build and run these tests, run:
+*
+
+ cargo test -F tap-tests --no-run --test tap_tests |& \
+           sed -e 's/[()]//g' | awk '/Executable/ {print $3}' | xargs sudo
+
+* which does the build as a non-priv user, extracts the exec binary location of
+* the test from 'cargo test', and runs only that as root.
+*
+* To develop on these tests, you need to manually enable this feature in VSCode ala:
+*
+*  In VS Code, open the Extensions sidebar, click the gear icon next
+*  to the rust-analyzer extension, and choose “Extension Settings.”
+*  You can choose whether to customize settings for all projects (the
+*  “User” tab) or just the current one (the “Workspace” tab). The
+*  setting is labeled “Cargo: Features.”
+*
+* [from https://users.rust-lang.org/t/passing-feature-flags-to-rust-analyzer/45918/3]
+*
+* to debug, run:
+*
+* sudo rust-gdb ./target/debug/deps/tap_test-${BUILD}
+*  break tap_test::tests::<TAB>   # to get a list of useful breakpoints
+*
+* NOTE: tests in rust capture stdio/stderr by default; add "-- --nocapture", e.g.,
+*  'cargo test -- --nocapture'
+*/
+#[cfg(feature = "tap-tests")]
+mod tests {
+
+    use etherparse::{PacketBuilder, PacketHeaders};
+    use pcap::Capture;
+    use tun_tap::Iface;
+
+    /***
+     * Create a Tap interface and make sure that the sendpacket() and next_packet()
+     * work as expected
+     */
+    #[test]
+    fn conntrack_tap_basic() {
+        let (cap, iface) = capture_tap_interface();
+
+        // NOTE: on Linux, if you don't specify a timeout(), it will never return
+        let mut cap = cap.snaplen(32000).timeout(500).open().unwrap();
+
+        // create a test packet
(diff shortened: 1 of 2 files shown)


tests/tap_tests.rs
```diff
@@ -0,0 +1,149 @@
+/***
+* These tests need to be run as root and currently only work on Linux (and maybe MacOs?)
+*
+* To build and run these tests, run:
+*
+
+ cargo test -F taptest --no-run --test tap_tests |&
```
stappersg, line 7: The ampsand at the end should probably be backslash, as an indicator of continueing line.

robs-zeynet: Thanks for the review.  The '|&' is intentional to capture stderr + stdout (this is a /bin/bash'ism), but I agree the backslash would help with readability.  Let me update.

Any other comments on the PR?

stappersg: Today I learnt that `|&` is shorthand for `2>&1 |`.

Wojtek242:
Thanks for the PR @robs-zeynet! I think this is a really valuable addition to the repository! Sorry for the late reply.

I would like to merge this contribution straight away, but CI is stopping me. It appears that due the use of the `--all-features` flag it's catching your new feature flag and running it as well. This doesn't work in the CI since it doesn't have root and you use features to new for the oldest rust version we support. However, the test is still valuable to run manually and as a reference and example for further tests of this sort so I don't see that as a problem.

Having said that, I would like to continue using the `--all-features` flag in CI. Do you see any other way of excluding this test when `--all-features` is enabled?

robs-zeynet:
Hello and thank you for the kind words.

One way to work around the `--all-features` issue  - perhaps a bit hacky - would be to create a second feature (maybe called "all-features"?) and wrap the test in an:

```#[cfg(all(feature = "tap-tests", not(feature = "all-features))]```

that way the code would only trigger when 'tap-tests' was enabled and ''all-features" was not.  Sound ok?

Wojtek242:
Uff, this is a tricky one. Some of the new dev-dependencies don't work with out MSRV. I've played around on my machine and what I think should work is the following:
1. Move `etherparse` and `tun-tap` to `[dependencies]` and mark them `optional = true` (you cannot mark `[dev-dependencies] optional).
2. Make the `tap-tests` feature depend on `etherparse` and `tun-tap`.

It's a shame that this won't work with `dev-dependencies`. I'll think about a better long-term solution, or perhaps just wait until it's time to bump the MSRV.

stappersg:
minimum supported Rust version (MSRV)

robs-zeynet:
> minimum supported Rust version (MSRV)

Admittedly I had to google it :-)

robs-zeynet:
> It's a shame that this won't work with `dev-dependencies`. I'll think about a better long-term solution, or perhaps just wait until it's time to bump the MSRV.

Agree that this isn't ideal (why can't you have optional dev dependencies!?) but your proposed workaround seems fine to me as both libraries are relatively small.  Shall I just make the change or is there a reason to not consider your workaround?

Wojtek242:
> > It's a shame that this won't work with `dev-dependencies`. I'll think about a better long-term solution, or perhaps just wait until it's time to bump the MSRV.
> 
> Agree that this isn't ideal (why can't you have optional dev dependencies!?) but your proposed workaround seems fine to me as both libraries are relatively small. Shall I just make the change or is there a reason to not consider your workaround?

Nah, just go for it. We don't publish the crate automatically so there will be time to clean it up if there is a need to clean it up before publishing the next version.

robs-zeynet:
Updated - feedback welcome!

Wojtek242:
Brilliant! Thanks, for the contribution!

PR #287, opened 2023-01-01
docs: fix broken leading example

amyipdev:
The first example listed in the documentation is broken and does not compile. This patch fixes the error, allowing it to compile correctly.

Signed-off-by: Amy Parker <apark0006@student.cerritos.edu>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,7 +12,7 @@
 //! use pcap::Device;
 //!
 //! fn main() {
-//!     let mut cap = Device::lookup().unwrap().open().unwrap();
+//!     let mut cap = Device::lookup().unwrap().unwrap().open();
 //!
 //!     while let Ok(packet) = cap.next_packet() {
 //!         println!("received packet! {:?}", packet);


stappersg:
> This patch fixes the error, allowing it to compile correctly.

On which Operating System is the error encountered?
With which tool chain is the _it now works for me_  is experienced?

amyipdev:
@stappersg All supported operating systems. The problem was that the first function returns a `Result<Option<Device>>`; the first unwrap gives an `Option<Device>`, which does not implement `.open()` as its an option. `Device` does implement `.open()`, so you need to unwrap again. I think whoever made the example just switched around the calls - compile the current one for yourself, it won't work, while the one I provided will. 

I am currently on Rust 1.66.0 `stable-x86_64-unknown-linux-gnu`, error encountered on Arch Linux with kernel=`6.1.1`, libpcap=`1.10.1`, pcap=`1.0.0`.

Wojtek242:
@amyipdev thanks for the PR!

You are indeed correct that an additional `unwrap` is needed before the `open`, but the `unwrap` after `open` is also needed as `open` returns a `Result`. If you add that `unwrap` I will merge this PR.

amyipdev:
Will try to add later today, if I don't get it done by end of day just send another message here and I'll get it in

Wojtek242:
Resolved by #296

PR #292, opened 2023-01-29
 clippy: variables can be used directly in the format! string

Wojtek242:
Allowing the lint instead of fixing as the oldest rust compiler supported by pcap does not support the new feature.


Stargateur:
> This lint is aware of the MSRV:
> 
> https://github.com/rust-lang/rust-clippy/blob/54e929bddee5934c527894fabee8a2c48531a795/clippy_lints/src/format_args.rs#L205

https://github.com/rust-lang/rust-clippy/issues/10087, Can we add msrv field that get a warning ?

PR #296, opened 2023-05-03
Fix the examples in the docs

Julian-Wollersberger:
And mark them as `no_run` instead of `ignore`, so they don't regress again.

Apparently the return type of `Device::lookup()` had changed and the docs weren't updated.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -8,16 +8,15 @@
 //! use `.open()` on a `Device`. You can obtain the "default" device using
 //! `Device::lookup()`, or you can obtain the device(s) you need via `Device::list()`.
 //!
-//! ```ignore
+//! ```no_run
 //! use pcap::Device;
 //!
-//! fn main() {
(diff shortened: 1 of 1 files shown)


Wojtek242:
Looks good to me. If you fix the clippy lints, I will merge.

Julian-Wollersberger:
Will do, but on Monday.

Julian-Wollersberger:
Clippy, cargo doc and cargo fmt should be happy now.

Wojtek242:
Something seems to have changed in beta over the weekend as this PR's checks are failing for the same reason the overnights suddenly failed today as well. If you tested clippy locally the PR should be fine. I'll merge once I figure out what's going on.

Julian-Wollersberger:
Yes, I tested with stable clippy locally. 
Just tested that `rustup toolchain install beta --component clippy --component rustfmt` invocation locally, and I get the same error as CI, so I guess the current beta is broken.

PR #302, opened 2023-06-22
Add example with saving packets to different files

maybe-hello-world:
This PR adds an example of using Capture::dead() to save packets from an active Capture to different files.

## Reason
Splitting the PCAP stream (either in-flight or from an existing file) is a common task, and currently, no example covers it. If a user tries to create `SaveFile` from the same active Capture, errors are not helping a lot (current errors are about mutable and immutable borrows, while the right approach - create new Captures with a `dead` parameter).

### DIFF
--- a/examples/savemultiplefiles.rs
+++ b/examples/savemultiplefiles.rs
@@ -1,0 +1,32 @@
+use pcap::Capture;
+
+fn main() {
+    // get the default Device
+    let device = pcap::Device::lookup().unwrap().unwrap();
+
+    // Setup Capture
+    let mut cap = pcap::Capture::from_device(device)
+        .unwrap()
+        .immediate_mode(true)
+        .open()
+        .unwrap();
+
+    // remember linktype to create PCAP files later
+    let linktype = cap.get_datalink();
+
+    // Save each 30 packets into a new PCAP file
+    let mut counter = 0;
+    loop {
+        let mut save_file = Capture::dead(linktype)
+            .unwrap()
+            .savefile(format!("dump_{}.pcap", counter))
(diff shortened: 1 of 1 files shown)


examples/savemultiplefiles.rs
```diff
@@ -0,0 +1,32 @@
...
+        .unwrap()
+        .immediate_mode(true)
+        .open()
+        .unwrap();
+
+    // remember linktype to create PCAP files later
+    let linktype = cap.get_datalink();
+
+    // Save each 30 packets into a new PCAP file
+    let mut counter = 0;
```
Stargateur, line 18: I would prefer 0usize

examples/savemultiplefiles.rs
```diff
@@ -0,0 +1,32 @@
...
+        let mut save_file = Capture::dead(linktype)
+            .unwrap()
+            .savefile(format!("dump_{}.pcap", counter))
+            .unwrap();
+
+        for _ in 0..30 {
+            let packet = cap.next_packet().unwrap();
+            save_file.write(&packet);
+        }
+
```
Stargateur, line 29: A call to flush would allow to check for write error

Stargateur (changes requested):
don't like much this infinite loop

maybe-hello-world:
Switched from infinite loop to just 5 files. You're correct, for this example's purposes it's fine, and most likely we don't want some example running forever.

Stargateur:
Now you can use the for loop count instead of doing your counter manually :p

maybe-hello-world:
Thanks, updated. :)

PR #305, opened 2023-08-05
Add example reading from stdin

mmun:
I recently had to do something like this and it wasn't immediately clear to me whether or not it was possible with this library. I figured it would help someone else if I added an example.


Wojtek242:
Thanks!

PR #311, opened 2023-10-30
Update windows-sys dependency

qrnch-jan:
`windows-sys` has been at `0.48` for some time now, and is pretty widely used.  I think we should upgrade to it.

There was one minor fallout; `sa_family` is a `u16`, but for some odd-ball reason it was `u32` in `0.36.1`.  (This was likely due to a now-fixed bug in the `win32metadata` project, which is used to generate the `windows-sys` crate).

No external interfaces are affected (apart from `HANDLE` being from `windows-sys` `0.48` rather than `0.36.1`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+### Changed
+
+- Crate maintenance: Use `windows-sys` version `0.48.0`.
+- Deprecate `open_raw_fd()`.
+
 ## [1.1.0] - 2023-05-12
 
 ### Added
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ tun-tap = { version = "0.1.3", optional = true }
 etherparse = { version = "0.13.0", optional = true }
 
 [target.'cfg(target_os = "windows")'.dependencies]
-windows-sys = { version = "0.36.1", features = ["Win32_Foundation", "Win32_Networking_WinSock"] }
+windows-sys = { version = "0.48.0", features = ["Win32_Foundation", "Win32_Networking_WinSock"] }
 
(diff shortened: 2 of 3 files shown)


Stargateur (commented):
I would mark this obsolete to remove it in the next major version, I don't even see why we expose this, if user want open a fd there is better way in rust.

Stargateur (approved):

qrnch-jan:
Are the failing tests actually failed tests?  It looks like it's timing out updating from crates.io?

Wojtek242:
Thanks for the PR! Much appreciated, just like all your previous ones :-)

> Are the failing tests actually failed tests? It looks like it's timing out updating from crates.io?

Yes, the CI pipeline has been broken for some time. I'm pretty sure it's due to some infra update by GitHub as the tests that do end up running tend to complete fine.

I haven't had the time or motivation just yet to fix them. I apologise, but the PR may have to wait until I (or somebody else) has a look and figures out what's wrong with the CI.

Stargateur:
plz just use std rust for open a file, https://doc.rust-lang.org/std/os/fd/trait.IntoRawFd.html#tymethod.into_raw_fd

qrnch-jan:
Just checking in to say I have not forgot about this and will return to it.  Swamped with $DAYJOB for now.

Wojtek242:
Due to significant changes on main since this was proposed, I will close it. If you get around to addressing the comments and adapting to the current code base, I'll be happy to review it.

Thought I suspect that the `windows-sys` update will be held back by MSRV. If it helps, we can bump MSRV to 1.48 as that's what Debian bullseye provides.

Wojtek242:
I have recently updated the MSRV to 1.63. This should make this PR easier to implement if you ever choose to come back to it.

PR #313, opened 2023-11-10
Fix complaints from a 2021 clippy

saethlin:
I was searching for discussion about this crate and happened to find someone who was confused by the implicit lifetimes. So I locally bumped the edition and looked at what `cargo clippy` complained about. They all seem like reasonable changes to me, and hopefully the explicit lifetimes help out future newcomers.

### DIFF
--- a/src/codec.rs
+++ b/src/codec.rs
@@ -8,5 +8,5 @@ use crate::Packet;
 pub trait PacketCodec {
     type Item;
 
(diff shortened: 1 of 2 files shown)


saethlin:
Yup, that's what I was afraid of. The release of `memchr` 2.6.0 contained an MSRV bump, which means that this crate, unless something is done, has also changed its MSRV.

Wojtek242:
Thanks a lot!

PR #317, opened 2023-11-11
Try to fix windows DLL issue

Wojtek242:


Wojtek242:
Gonna resolve by testing against mocks instead.

PR #320, opened 2023-12-11
add Windows support for capture-stream

### LINKED ISSUE #191: Windows support for `capture-stream` feature
I'm currently using the `pcap-async` crate to perform packet capturing in asynchronous code.
I'm thinking about switching to this crate due to the Windows support via Npcap/WinPcap. But you write that the `capture-stream` feature is only supported on Linux and macOS.

What is the reason for that? Can this be easily fixed?

Freax13:
This PR adds Windows support for capture-stream.

Closes #191

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,10 +21,6 @@ jobs:
         toolchain: [stable, beta]
         profile: ['', --release]
         features: ['', '--all-features']
-        exclude:
-          # capture-stream is not supported on Windows.
-          - os: windows-latest
-            features: '--all-features'
 
     uses: ./.github/workflows/ci-steps.yml
     with:
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,13 +19,17 @@ errno = "0.2"
 tokio = { version = "1.0", features = ["net", "rt", "macros", "rt-multi-thread"], optional = true }
 futures = { version = "0.3", optional = true }
 gat-std = { version = "0.1.1", optional = true }
-# these really belong in [dev-dependencies], but can't have optional dev-deps
-# and these libraries would truncate the min support Rust version (MSRV)
-tun-tap = { version = "0.1.3", optional = true }
+# This really belongs in [dev-dependencies], but we can't have optional dev-deps
+# and this libraries would truncate the min support Rust version (MSRV)
 etherparse = { version = "0.13.0", optional = true }
 
 [target.'cfg(target_os = "windows")'.dependencies]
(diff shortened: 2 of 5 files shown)


ColinFinck:
I really hope we can see this in a release soon!

CC @Wojtek242

Wojtek242:
Thanks for the PR @Freax13! I appreciate it and apologies for the delay in responding. I only check in every few months given where this project is at.

This looks like a good PR and would ultimately want to see it merged. The CI however fails as the newer windows-sys crate version no longer works with our MSRV. 

With that said, the stream feature was never required to work on the MSRV due to there being no chance for that ever happening with a dependency on tokio. Do you think it would be possible to hide the newer windows-sys dependency behind the stream feature flag? That way pcap with no features would still work on MSRV and the stream feature could work on a newer rustc version.

Wojtek242:
I would also appreciate some tests for the new feature. I don't work on Windows so I can't even test this manually. I would be willing to relax the MSRV if you could instead provide some unit tests for this MR.

Freax13:
> Thanks for the PR @Freax13! I appreciate it and apologies for the delay in responding. I only check in every few months given where this project is at.
> 
> This looks like a good PR and would ultimately want to see it merged. The CI however fails as the newer windows-sys crate version no longer works with our MSRV.
> 
> With that said, the stream feature was never required to work on the MSRV due to there being no chance for that ever happening with a dependency on tokio. Do you think it would be possible to hide the newer windows-sys dependency behind the stream feature flag? That way pcap with no features would still work on MSRV and the stream feature could work on a newer rustc version.

I reverted the windows-sys bump and inlined the one constant (`INFINITE`) I needed from the newer version.

> I would also appreciate some tests for the new feature. I don't work on Windows so I can't even test this manually. I would be willing to relax the MSRV if you could instead provide some unit tests for this MR.

While I understand the desire for unit tests, I'm not sure how we'd implement them. My main concerns are false positive errors with missing permissions and that I'm not sure how we could generate traffic on a known device. I assume those are also the reasons, this crate currently doesn't have unit tests for active devices (except for `tap_tests.rs`, but AFAICT those tests aren't run in CI).

Freax13:
1f53e0c419c6e70aac37a1ef9a8531845b96a418 broke CI on Windows. This should be fixed by a99e9ccb0a199459261ff79ef16b110fa44487a7.

Wojtek242:
Brilliant, thanks for the PR and addressing the MSRV issues! I can't test if this works, but I trust you guys tested it.

I will push to crates.io tonight or within the next few days.

PR #339, opened 2024-03-24
Try and figure out what's wrong with windows

Wojtek242:


Wojtek242:
Appears to be working on nightly so no need to stress

PR #346, opened 2024-04-17
Update libloading from 0.6 to 0.8

### LINKED ISSUE #345: Libloading declared most of its API unsafe in a subsequent release
This project depends on `libloading = 0.6`:

https://github.com/rust-pcap/pcap/blob/8b7bcfe12c79955c228e5657a1e63cdcce0af3f5/Cargo.toml#L43

It’s used in exactly one place:

https://github.com/rust-pcap/pcap/blob/8b7bcfe12c79955c228e5657a1e63cdcce0af3f5/build.rs#L102

But in the subsequent 0.7.0 release, `libloading` realized that most of its API, including `Library::new`, [should have been `unsafe`](https://github.com/nagisa/rust_libloading/blob/dc8664fd23a34b16e240c1bd48acbe3596cb8eef/src/changelog.rs#L98-L102). (The current release is 0.8.3)

“Hiding” soundness issues by using an old `libloading` version would seem to have a limited impact since only `build.rs` is affected, but it would be better to upgrade to a current version and actually deal with the `unsafe` API.

musicinmybrain:
Deal with the fact that `libloading::Library::new` was declared `unsafe` in `libloading 0.7`; see https://github.com/nagisa/rust_libloading/issues/86 and
https://github.com/nagisa/rust_libloading/blob/0.8.3/src/changelog.rs#L96-L151 for details.

Fixes #345.


Wojtek242:
Whilst I don't expect all CI jobs to pass at the moment, the Windows MSRV should pass, but it does not here: https://github.com/rust-pcap/pcap/actions/runs/8724374552/job/23943643474?pr=346. It appears that `libloading` introduced a dependency that requires a newer rust compiler.

Here's what you can do:
1. See if it will work with MSRV 1.63 which is Debian's rustc version. You can grep for `msrv` to see where you can change the CI to this version. You can also test older rust versions locally with rustup.
2. If the above doesn't work, perhaps a version of `libloading` newer than `0.6`, but older than the latest one will work.

musicinmybrain:
Thanks. I will investigate.

musicinmybrain:
Since `libloading` [cites an MSRV of 1.56 as of 0.8.2](https://github.com/nagisa/rust_libloading/blob/dc8664fd23a34b16e240c1bd48acbe3596cb8eef/src/changelog.rs#L15), I’m trying that first. Hopefully I’ve tweaked everything correctly.

musicinmybrain:
It looks like this now has no more CI failures than https://github.com/rust-pcap/pcap/pull/348 does.

Wojtek242:
> It looks like this now has no more CI failures than #348 does.

Yes, the CI currently fails in the same places as main. I will merge and publish soon. Thanks for the PR!

musicinmybrain:
You’re welcome! Glad it is useful.

PR #348, opened 2024-04-18
Fix IPv4 initialization for test_address_ipv4

### LINKED ISSUE #347: On big-endian systems, test_address_ipv4 fails
I’m testing an update of the `rust-pcap` package in Fedora Linux to 1.3.0, and I’m finding that one test fails on the `s390x` architecture, which is big-endian.

```
---- device::tests::test_address_ipv4 stdout ----
thread 'device::tests::test_address_ipv4' panicked at src/device.rs:644:9:
assertion `left == right` failed
  left: "66.0.0.10"
 right: "10.0.0.66"


failures:
    device::tests::test_address_ipv4
```

Clearly, the byte order is getting reversed at some point.

It looks like the `set_addr()` methods for IPv4 are designed to accept a `u32` in network byte order, as the value is passed directly through to the underlying structure:

https://github.com/rust-pcap/pcap/blob/8b7bcfe12c79955c228e5657a1e63cdcce0af3f5/src/device.rs#L365-L367
https://github.com/rust-pcap/pcap/blob/8b7bcfe12c79955c228e5657a1e63cdcce0af3f5/src/device.rs#L379-L381

But in `fn sockaddr_ipv4()`, the default address is assigned from an integer literal whose byte representation depends on host endianness, and which has been chosen to have the expected value on little-endian systems:

https://github.com/rust-pcap/pcap/blob/8b7bcfe12c79955c228e5657a1e63cdcce0af3f5/src/device.rs#L391

musicinmybrain:
The address used in the test was incorrectly initialized on big-endian hosts.

Fixes #347.

I tested this as a patch using real `s390x` hardware.

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -388,7 +388,7 @@ mod tests {
         let mut addr: WinSock::SOCKADDR_IN = InetAddressV4::new();
 
         addr.sin_port = 1075;
-        addr.set_addr(0x4200000A);
+        addr.set_addr(0x0A000042_u32.to_be());
 
         Sockaddr::SockaddrIn(addr)
     }


Stargateur (approved):

Stargateur:
In perfect world I would prefer a better API, probably set_addr should do the to_be() but it's only for testing soooooo.

Wojtek242:
Thanks for the PR! The CI doesn't pass, but the failures are the same as on main (I just haven't gotten around to resolving them). I'll handle it and merge soon.

musicinmybrain:
You’re welcome! Thanks for reviewing it.

Issue #340, opened 2024-03-24
Windows test fail with `exit code: 0xc0000142, STATUS_DLL_INIT_FAILED`


Wojtek242:
For now only on the `beta` channel. The exact same code works on `stable` and `msrv`. As usual, it is intuitive to blame npcap, but the unit tests are designed to run without npcap and they've been working so far.

Wojtek242:
https://github.com/rust-pcap/pcap/actions/runs/8407408622/job/23022345594

Wojtek242:
Some hope for 1.78/1.79? https://github.com/rust-pcap/pcap/actions/runs/8775072751/job/24076974369

PR #356, opened 2024-05-03
Run all tests on Windows

### LINKED ISSUE #275: Intermittent Windows CI failures

Wojtek242:
Closes #275, #340

### DIFF
--- a/.github/workflows/00-windows.yml
+++ b/.github/workflows/00-windows.yml
@@ -32,6 +32,9 @@ jobs:
     uses: './.github/workflows/02-coverage.yml'
     with:
       os: 'windows-latest'
+    secrets:
+      NPCAP_OEM_PASSWORD: ${{ secrets.NPCAP_OEM_PASSWORD }}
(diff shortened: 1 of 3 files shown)


Wojtek242:
Try again when 1.79 comes out on 13 June

PR #366, opened 2024-08-27
Add `want_pktap` on `Capture<Inactive>`

MarkusPettersson98:
This PR adds a binding for the macOS-specific `pcap_set_want_pktap` and exposes it on the `Capture<Inactive>` state. This function is not that well documented, but it exists in Apple's fork of `libpcap` since version `1.5.3`: https://opensource.apple.com/source/libpcap/libpcap-48/libpcap/pcap-bpf.c.auto.html

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -29,6 +29,7 @@ impl Version {
         vec![
             Version::new(1, 2, 1),
             Version::new(1, 5, 0),
+            Version::new(1, 5, 3),
             Version::new(1, 7, 2),
             Version::new(1, 9, 0),
             Version::new(1, 9, 1),
--- a/src/capture/inactive.rs
+++ b/src/capture/inactive.rs
@@ -98,6 +98,14 @@ impl Capture<Inactive> {
         self
     }
 
+    /// Set want_pktap to true or false. The default is maintained by libpcap.
+    #[cfg(all(libpcap_1_5_3, target_os = "macos"))]
+    pub fn want_pktap(self, to: bool) -> Capture<Inactive> {
+        unsafe { raw::pcap_set_want_pktap(self.handle.as_ptr(), to as _) };
+
+        self
+    }
+
     /// Set rfmon mode on or off. The default is maintained by pcap.
     #[cfg(not(windows))]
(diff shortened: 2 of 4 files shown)


Wojtek242:
Thanks for your contribution! Overall looks good to me and I'd be happy to merge it and release it.

However, the CI is failing. Could you please write the missing unit test (which I expect to be trivial, but it makes my life easier if the CI can just check for 100% coverage) and fix the lint issues?

MarkusPettersson98:
Thanks for the quick reply! I just noticed the CI issues as well, will fix them before marking this PR as ready for review again. Sorry for the inconvenience :)

src/capture/inactive.rs
```diff
@@ -98,6 +98,17 @@ impl Capture<Inactive> {
         self
     }
 
+
+    /// Set want_pktap to true or false. The default is maintained by pcap.
```
Wojtek242, line 102: The default is maintained by **libpcap**. We use a naming convention where `libpcap` refers to the underlying library and `pcap` refers to this crate.

build.rs
```diff
@@ -29,6 +29,7 @@ impl Version {
         vec![
             Version::new(1, 2, 1),
             Version::new(1, 5, 0),
+            Version::new(1, 5, 3),
```
Wojtek242, line 32: Please also add a `1.5.3` step to the CI in https://github.com/rust-pcap/pcap/blob/main/.github/workflows/01-build-and-test-unix.yml.

Wojtek242 (changes requested):
And some very minor changes.

build.rs
MarkusPettersson98: Done

src/capture/inactive.rs
MarkusPettersson98: Thanks for catching that, updated the comment :)

Wojtek242 (approved):

Wojtek242:
Brilliant, thanks!

MarkusPettersson98:
Thanks for your guidance and quick response! :purple_heart:

PR #367, opened 2024-08-30
Implement AsFd for Capture<Active>

gmacon:
I wanted to use [`nix::poll::poll`](https://docs.rs/nix/latest/nix/poll/fn.poll.html) to wait for the next packet, and a `BorrowedFd` is needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Added an implementation of `AsFd` on `Capture<Active>` on non-Windows.
+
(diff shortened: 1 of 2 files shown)


Wojtek242:
Thanks for the PR! Looks good to me as it is. Can you just update the CHANGELOG as well? I will then release it shortly after I merge it.

gmacon:
Thank you for turning this around so fast. I had implemented equivalent functionality as a free function in anticipation of having to wait for this to end up in a release, but you got the release done before my colleagues got my changes reviewed, so I've just removed the free function and called this one instead. I really appreciate it.

PR #372, opened 2024-11-19
Use normal stream terminating mechanism

jethrogb:
`Stream`, like `Iterator`, returns an `Option` to indicate end-of-stream with `None`. Implement this.


Wojtek242:
@jethrogb Thanks for the PR!

Can I ask you to add a note to the CHANGELOG before I merge?

PR #373, opened 2024-11-19
Add async sink for sending packets

jethrogb:
Add an implementation of `futures::Sink` to be able to send packets asynchronously. (UNIX only - I'm not too familiar with Windows).

The `Sink` API is a bit difficult to understand. This implementation is modeled after [`Unfold`](https://docs.rs/futures/0.3.31/futures/sink/struct.Unfold.html) and [`IntoSink`](https://docs.rs/futures/0.3.31/futures/io/struct.IntoSink.html).

### DIFF
--- a/src/sink/mod.rs
+++ b/src/sink/mod.rs
@@ -1,0 +1,118 @@
+//! Support for asynchronous packet transmission.
+//!
+//! See [`Capture::sink`](super::Capture::sink).
+use std::io;
+use std::marker::Unpin;
+use std::pin::Pin;
+use std::task::{self, Poll};
+
+use futures::ready;
+use tokio::io::unix::AsyncFd;
+
+use crate::{
+    capture::{selectable::SelectableCapture, Active, Capture},
+    Error,
+};
+
+impl Capture<Active> {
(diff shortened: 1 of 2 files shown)


src/sink/mod.rs
```diff
@@ -0,0 +1,118 @@
...
+        cx: &mut task::Context<'_>,
+    ) -> Poll<Result<(), Self::Error>> {
+        let sink = Pin::into_inner(self);
+        let packet_storage = &mut sink.packet;
+
+        loop {
+            let mut guard = ready!(sink.inner.poll_write_ready_mut(cx))?;
+
+            if let Some(packet) = packet_storage.take() {
+                match guard.try_io(
```
raoulstrackx, line 86: Nitpick: this `match` is a bit hard to parse due to the closure. Consider create a separate variable for it.

raoulstrackx (approved):

Wojtek242:
@jethrogb Thanks for the PR!

Could you please do the following so that I can merge this PR:
1. Add a CHANGELOG entry,
2. Fix clippy lints
3. Add unit tests

PR #375, opened 2024-12-16
Make docs.rs source view generate links to definitions.

qrnch-jan:
When browsing the source on docs.rs it's nice to be able to click on items to reach their definitions.  This is supported, but is opt-in (which this patch does).  (See https://docs.rs/ump/latest/src/ump/lib.rs.html#111 for an example of what this can look like).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ mockall = "0.11.4"
 tempfile = "3.10"
 
 [target.'cfg(target_os = "windows")'.dev-dependencies]
-eui48 = "1.1"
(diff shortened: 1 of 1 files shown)


Wojtek242:
@qrnch-jan Thanks for the patch! That looks like a pretty neat feature.

@stappersg Thank you for your comment. Whilst your remark is technically sound, in my opinion it unnecessarily overburdens an active contributor in a very quiet repository. Could I please ask you to leave only reviews that address functional issues that improve functionality rather than just code/commits.

PR #376, opened 2024-12-17
Add an example illustrating how to use event semaphore on Windows and…

qrnch-jan:
… update `Capture::get_event()` documentation to declare the intended use of the event semaphore.

The example is a little larger than usual, but I wanted to show not only the actual semaphore wake-up, but also how it could be paired with a simple signalling mechanism as well.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ tempfile = "3.10"
 
 [target.'cfg(target_os = "windows")'.dev-dependencies]
 eui48 = "1.1"
+windows-sys = { version = "0.36.1", features = ["Win32_System_Threading"] }
 
 [target.'cfg(not(target_os = "windows"))'.dev-dependencies]
 tun-tap = "0.1.3"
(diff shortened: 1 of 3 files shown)


stappersg:
Please convert the _merge request text_ into _commit message_.

qrnch-jan:
> Please convert the _merge request text_ into _commit message_.

I very rarely use github, or even git for that matter, so I'm going to need some guidance.  I understand your principle message:  "We do not want github's metadata to track history for us, we want all the journal entries to be in the actual repo data", and that I can fully appreciate and will make sure to do it that way in the future.

That said, I'm unclear on how to amend this for the existing PR's. Should I `git commit --amend` _on the original PR commit id_ and fix the commit message there, or should it be done on the latest commit?  Or is there some other way to do it?

Wojtek242:
@qrnch-jan and @stappersg  There is no need to change the commit message because it can be ammeded during the squash and merge and if necessary I will do so when accepting the PR.

@qrnch-jan Thanks for the new example!

PR #377, opened 2024-12-17
Disable default features for eui48 to avoid pulling in rustc-serialize.

qrnch-jan:
rustc-serialize has a history of causing builds to output annoying deprecation warnings.  Explicitly make sure that won't happen in the future by disabling all default features from eui48 crate.


stappersg (commented):
The patch looks good, the commit message could (should???) be better.

Wojtek242:
@qrnch-jan thanks for the PR! I'll just merge it straight ahead since it seems all checks pass.

@stappersg I think the commit message is clear enough.

PR #379, opened 2024-12-18
Sendqueue pkt header size

qrnch-jan:
There's currently a gap for applications that want to precalculate the exact `SendQueue` size required for a batch of packets -- the application doesn't know[*] the exact size of the packet header that is implicitly added with each packet.  This PR adds a constant function, tied to `SendQueue`, that will allow applications to easily get the size without bothering with any other details.

[*] Well, this isn't technically true -- it happens to be the same size and layout as packet headers found elsewhere in the crate.  However, the benefit of this implementation is that it considers the exact type an implementation detail that the application doesn't need to care about.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,18 @@
 
 ## [Unreleased]
 
+### Added
+
+- Added `packet_header_size()` for applications that need to know the size of internal type
+  `pcap_pkthdr` (for instance to calculate exact send queue sizes).
+
+want to precalculate exact queue sizes.
+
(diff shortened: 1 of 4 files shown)


qrnch-jan:
@Wojtek242 It is indeed not strictly Windows-exclusive, but the reason I put it there was to emphasize that it's the size of the packet header used by {win,n}pcap send queues, and it's the size one should use to calculate required sendqueue sizes.

I have moved the function, and documented that it's the one that should be used by `SendQueue`.

Wojtek242 (approved):

PR #380, opened 2024-12-18
Docsrs labels

qrnch-jan:
rustdoc/docs.rs support adding cfg/feature labels to generated docs, but it's currently nightly only and put behind a feature gate.  Apparently there are some future goals to make generation of these automatic, but for now it requires adding the entries manually.  docs.rs happens to use nightly to generate docs, so we can make use of that while retaining compatibility with stable rustdoc.

With these changes, `cargo doc` will work the same as it always has, but following the added instructions in the readme will yield labels like the ones (for instance) tokio has: https://docs.rs/tokio/1.42.0/tokio/#modules

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,7 +31,7 @@ mockall = "0.11.4"
 tempfile = "3.10"
 
 [target.'cfg(target_os = "windows")'.dev-dependencies]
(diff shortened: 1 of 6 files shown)


Wojtek242:
LGTM - thanks for the suggestion!

PR #391, opened 2025-11-04
Add a binding for pcap_breakloop

Stargateur:
Alternative to https://github.com/rust-pcap/pcap/pull/390

Didn't test cause I'm on windows right now. But test is doubt anyway.


Wojtek242:
Thanks @Stargateur. However, I would first like to give @cai-william a chance to amend his PR based on your suggestions so that GitHub acknowledges him as a contributor.

PR #392, opened 2025-11-05
Don't automatically link to libpcap

marshallpierce:
pkg-config's default config will output `rustc-link-lib` when it finds a package. However, per this project's readme, linkage is the responsibility of the user, so it seems like merely probing for the version should not change any subsequent build config.

Lines previously present in build.rs output that are now gone:

```
[pcap 2.3.0] cargo:rustc-link-search=native=/usr/lib/x86_64-linux-gnu
[pcap 2.3.0] cargo:rustc-link-lib=pcap
```

Similarly, the `#[link(name = "pcap")]` in `raw` was making dependent binaries automatically link to libpcap.so.

Dependent crates will need to either use a `#[link...` snippet, or specify linkage via [build.rs](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-link-lib). That matches my understanding of the readme's linkage policy, and for what it's worth, as a user of the crate I'm fine with that. A downside with this approach is that it's possible to link `libpcap` twice since this crate doesn't specify `link` in its Cargo.toml, but presumably that's something you've already considered as a worthwhile tradeoff to avoid needing to have ever more environment variables to configure linkage.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -197,6 +197,7 @@ fn main() {
 
 fn from_pkg_config() -> Result<pkg_config::Library, pkg_config::Error> {
     let mut config = pkg_config::Config::new();
+    config.cargo_metadata(false);
     // If the user has went out of their way to specify LIBPCAP_VER (even though
     // LIBCAP_LIBDIR wasn't set), respect it. Otherwise fall back to any version
     // as long as it's supported.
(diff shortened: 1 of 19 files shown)


Wojtek242:
Actually, I just noticed that you already have those flags - is it the case that even with those flags, the dynamic library is linked?

PR #399, opened 2026-08-01
Add a binding for pcap_dispatch

### LINKED ISSUE #388: Packet loss in high traffic environments
Hi guys, I recently created a [software](https://github.com/rikonaka/xxpdump-rs) that uses `libpnet` and `libpcap` to capture traffic. However, I found that `pcap` crate would lose some packets in high-traffic and high-load situations (such as nmap scans).

Here are some key codes.

https://github.com/rikonaka/xxpdump-rs/blob/de17d8fa07782bb70759a0fba358fb27fb77a99f/src/local.rs#L82-L100

https://github.com/rikonaka/xxpdump-rs/blob/de17d8fa07782bb70759a0fba358fb27fb77a99f/src/local.rs#L141-L148

I have set `snaplen` and `buffsize` to be large enough, such as `snaplen` to `65535` and `buffsize` to `16MB (163840)`. Even so, some packets are still lost when I receive data.

To avoid network and operating system problems, I used `tshark` and `tcpdump` to capture packets on my local machine. The results of both software included lost packets.

For example, this data packet below is the traffic of port 8080 (open) using `nmap` to perform SYN detection on all ports of the target (`sudo nmap -sS -p 22-65535 192.168.5.152`). When the filter is `tcp.port == 8080`, it should show a SYN detection packet from 192.168.5.3, an ACK packet from 192.168.5.152, and finally a RST packet from 192.168.5.3. Except for the RST packet, the other two packets are lost.

Image

Because the [software](https://github.com/rikonaka/xxpdump-rs) I wrote is a traffic capture software, I did not set any filters, and this problem occurs sporadically, sometimes the lost data is captured, and sometimes not. And the loss of data packets is random.

cicku:
Closes #388

### DIFF
--- a/src/capture/activated/mod.rs
+++ b/src/capture/activated/mod.rs
@@ -242,7 +242,8 @@ impl<T: Activated + ?Sized> Capture<T> {
     ///
     /// Unlike [`Self::next_packet()`], which performs a system call for every packet, this
     /// processes a whole buffer of packets in a single call, making packet loss less likely
-    /// in high traffic situations.
+    /// in high traffic situations. Packets are still dropped when the buffer fills up between
+    /// calls, so a busy interface also wants a larger [`Capture::buffer_size()`].
     ///
     /// At most `count` packets are processed. `None` processes all the packets received in
     /// one buffer when reading a live capture, or all the packets in the file when reading
@@ -250,7 +251,9 @@ impl<T: Activated + ?Sized> Capture<T> {
     ///
     /// Unlike [`Self::for_each()`], this does not block until `count` packets have been
     /// processed: it returns the number of packets processed as soon as one buffer has been
-    /// handled, which may be zero if the timeout expired or there are no more packets to read.
+    /// handled, which may be zero if there are no more packets to read. Note that on most
+    /// platforms the read timeout only starts once the first packet arrives, so a quiet live
+    /// capture can block instead of returning zero.
     pub fn dispatch<F>(&mut self, count: Option<usize>, handler: F) -> Result<usize, Error>
     where
         F: FnMut(Packet),
@@ -281,8 +284,10 @@ impl<T: Activated + ?Sized> Capture<T> {
         if let Some(e) = handler.panic_payload {
             resume_unwind(e);
         }
+        // A successful pcap_dispatch returns the number of packets processed, not 0 like
+        // pcap_loop.
         self.check_err(return_code >= 0)
-            .map(|_| return_code as usize)
+            .and(Ok(return_code as usize))
     }
 
     /// Returns a thread-safe `BreakLoop` handle for calling pcap_breakloop() on an active capture.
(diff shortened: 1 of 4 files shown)


RazorBest:
Hi cicku! Thanks for submitting this PR! Your effort is appreciated!

## Testing

I tried testing this starting from @rikonaka's demo, and changed it: https://github.com/RazorBest/pcap-issues

There are 3 scenarios:
- pcap-rs with the old `next_packet` (implemented in `src/main.rs::pcap_test_old`)
- pcap-rs with your new `dispatch` (implemented in `src/main.rs::pcap_test_new`)
- C version that uses `pcap_dispatch` (in `sniffer.c`)

The test environment assumes:
- sudo access to both C and rust sniffers
- a Docker container running on `172.17.0.2`, accessible through the `docker0` interface
- the host can send packets to `172.17.0.2`

To run the Docker container:
```
docker build -f Dockerfile_echo . -t echo
docker run -ti --rm echo
```

To run the packet capture:
```
cargo run # Change src/main.rs depending on which scenario you want
```
Or, for the C version:
```
gcc sniffer.c -lpcap && sudo ./a.out
```

While the container is up and the sniffer is capturing, run:
```
sudo nmap --max-rate 2000 -sS -p 22-8192 172.17.0.2
```

Then, hit `Ctrl+C` in the packet capture terminal. This doesn't immediately stop the process. An additional packet needs to be received. You can run `nc 172.17.0.2 8080` to trigger that.

## Result

~With your PR, there are still missing ports:~

~With sniffer.c, there are no missing ports. This indicates that `pcap_dispatch` is able to catch all the packets. So the issue might be somewhere in the Rust implementation.~

Edit: Actually, there was a difference between the C and Rust version. The C version was using a bigger buffer. Now, that the Rust version has a big enough buffer, your PR passes.

src/capture/activated/mod.rs
```diff
@@ -1215,4 +1264,184 @@ mod tests {
...
+        let test_capture = test_capture::<Active>(pcap);
+        let mut capture: Capture<dyn Activated> = test_capture.capture.into();
+
+        let mut packets = 0;
+        let processed = capture
+            .dispatch(Some(0), |_| {
+                packets += 1;
+            })
+            .unwrap();
+        assert_eq!(packets, 0);
```
RazorBest, line 1499: This test is a little shallow. The context doesn't generate any packet, so even if you run dispatch with a higher count, packets would still be 0. I'd provide the same context defined in `dispatch_with_count`.

src/capture/activated/mod.rs
```diff
@@ -1215,4 +1264,184 @@ mod tests {
...
+        assert_eq!(processed, 0);
+    }
+
+    #[test]
+    fn dispatch_timeout_expired() {
```
RazorBest: I don't see how this is testing the timeout. From my tests, it seemed that dispatch was not returning unless a packet was captured.

src/capture/activated/mod.rs
```diff
@@ -1215,4 +1264,184 @@ mod tests {
...
+            .withf_st(move |arg1| *arg1 == pcap)
+            .return_once_st(move |_| {});
+
+        capture
+            .dispatch(None, |_| panic!("panic in callback"))
+            .unwrap();
+    }
+
+    #[test]
+    fn dispatch_with_count() {
```
RazorBest, line 1350: Maybe there would be nice to have two other tests? One where the count provided to dispatch is higher than the amount of issued packets, and one where it's lower.

src/capture/activated/mod.rs
```diff
@@ -236,6 +238,53 @@ impl<T: Activated + ?Sized> Capture<T> {
...
+                cnt,
+                HandlerFn::<F>::callback,
+                &mut handler as *mut HandlerFn<AssertUnwindSafe<F>> as *mut u8,
+            )
+        };
+        if let Some(e) = handler.panic_payload {
+            resume_unwind(e);
+        }
+        self.check_err(return_code >= 0)
+            .map(|_| return_code as usize)
```
RazorBest: In `for_each`, we do `self.check_err(return_code == 0)`. I'd vouch for consistency, even if this code is correct.

RazorBest (commented):
Reviewed mod.rs.

src/capture/activated/mod.rs
cicku: Name changed.

src/capture/activated/mod.rs
cicku: Added 2 more

src/capture/activated/mod.rs
cicku: Added 2 more

src/capture/activated/mod.rs
cicku: `pcap_loop` returns 0 on success, but `pcap_dispatch` returns the number of packets processed, so `== 0` would treat every successful call that captured a packet as an error.

cicku:
@Wojtek242 IMO it is ready to be merged, please take a look when you have time.

RazorBest (commented):
Ran the tests and looked again at the tests in mod.rs. They seem to do what they "advertise".

Wojtek242:
@cicku and @RazorBest, thanks a lot for your work! I'll have a look when I'm back from holidays after the coming week.

Wojtek242:
Thanks @cicku for the PR and @RazorBest for the review and testing! I had a look at the code and it looks good to me and it passes the tests so will merge it and publish on crates ASAP.

PR #400, opened 2026-08-17
[3.0.0 prep] Base: Update dependencies as far as MSRV 1.85 allows

cicku:
bitflags 1.3 -> 2.13
errno 0.2 -> 0.3
windows-sys 0.36 -> 0.60
etherparse 0.13 -> 0.15
mockall 0.11 -> 0.12

---

Supersedes https://github.com/rust-pcap/pcap/pull/378. However, the version could either be 2.5.1, 2.6.0 or 3.0.0, I will leave it to the project maintainer.


Wojtek242:
Thanks for this! I think many windows users will appreciate the update.

Having carefully considered this, I would like to bump the major version to 3.0.0 as I would consider this a breaking change. Many have tried this windows dependency update and I can't do it myself since I don't use windows so I think it is worth doing it.

Since I will be be bumping the major version, I will also do two more things:
- I will bump the rust edition to 2024
- I will bump MSRV to 1.85
  - For the MSRV I usually guide myself with what does Debian ship and [trixie ships 1.85](https://packages.debian.org/search?keywords=rustc). Bookworm is technically still LTS but since it ships 1.63 and I had to bump MSRV to 1.64 for reasons I no longer remember then it's not relevant.
  - Rust edition 2024 also required 1.85

Therefore, if you would like to be bolder with the dependency updates, you are welcome to do so up to MSRV 1.85. If you've had enough that's also okay. I'll take what you've done so far and ship the new update.

What do you think?

Wojtek242:
I also see you've set your eye on a few other potential PRs. If there will be a major version bump now would be the time to do them as you will have some freedom to break a few things if they make your life easier (without make the changes too cumbersome). I would encourage you to submit those PRs. I will then wait until you say you're done with PRs so there would be no rush.

cicku:
Yes - this PR doesn't require immediate merge if we target 3.x. I will rebase accordingly and get back to you soon.

Wojtek242:
@cicku I see you marked this PR as ready. Just to check, is it ready for review?

cicku:
Yes. This one is ready, other PRs are also for 3.0.0 but still in progress

src/stream/windows.rs
```diff
@@ -64,15 +64,32 @@ impl<T: Activated + ?Sized, C: PacketCodec> futures::Stream for PacketStream<T,
...
 
-/// A wrapper around a HANDLE that can be used to call `WaitForSingleObject`
-/// from an asynchronous context. Once the call to `WaitForSingleObject`
-/// completes, the handle is considered ready and will keep returning `Ready`
-/// until it's reset.
+/// Drives a call to `WaitForSingleObject` on the capture's event from an
+/// asynchronous context. Once the call completes, the event is considered
+/// ready and will keep returning `Ready` until it's reset.
 struct EventHandle {
-    handle: HANDLE,
```
Wojtek242, line 72: To summarise, it looks like you moved the `handle` from inside `EventHandle` such that now when calling `poll_ready` you must also pass the `capture`.

In general, I strongly prefer such constructions where you only obtain handles at the moment they are necessary but since this will break user code (which is fine since the version is going up) could you explain the rationale for this change? What I'm looking for is: is the change worth the users whose code this will break updating it?

Wojtek242 (commented):
LGTM, I left one question just to understand your reasoning.

src/stream/windows.rs
cicku: This is more like a `Send` fix forced by the windows-sys bump. In windows-sys 0.36, `HANDLE` is isize; in 0.61 it's `*mut c_void`, so `PacketStream` picked up a raw pointer and stopped being `Send`.

---

Fetching the handle at the point of use for two reasons:

1. `PacketStream` then contains no raw pointer at all, so it's `Send` by ordinary auto-derivation. With a cached `SendHandle`, Send-ness rests on an unsafe impl covering a value held for the whole life of the stream, which a reviewer has to audit.
2. It drops the old SAFETY note's reliance on struct field ordering: _"PacketStream stores the handle before the capture, so the handle will be dropped before the capture"_.

---

IMO this change doesn't break any user code. `mod stream` is private in lib.rs and only `PacketStream` is re-exported, so `EventHandle` isn't nameable outside the crate and `PacketStream`'s public surface is unchanged. The user-visible break in this PR is `Capture::get_event()`'s return type, which windows-sys forces either way and is called out in the CHANGELOG.

Wojtek242: Sounds good to me

PR #401, opened 2026-08-17
Move the UN*X only entry points out of the shared FFI block

cicku:
This is a cosmetic change on top of libpcap 1.10.x.


Wojtek242:
Thanks! Looks pretty simple. Based on the comment I assume the change should not break windows builds.

Can you do one more thing before I merge? The unix workflow tests all the build options: https://github.com/rust-pcap/pcap/blob/main/.github/workflows/01-build-and-test-unix.yml - can you add an entry for 1.10.0 since you added to build.rs?

cicku:
Done. This can be merged safely before 3.x.

PR #402, opened 2026-08-21
[3.0.0 prep] tun_tap replacement 

cicku:
tun-tap (maintenance) pulls tokio-core, and with it the whole 0.1 era async stack, to open a file descriptor and read from it. tun-rs does the same job with no runtime and no cc, which also means the tap tests no longer need a C compiler to build.

This also shrinks the lock file.

Must be merged after #400, and a rebase will be done then.


cicku:
Ready: This PR only touches a test.

Wojtek242:
LGTM

PR #403, opened 2026-08-24
[3.0.0 prep] New pcap_* bindings

### LINKED ISSUE #383: [Bug] reading pcap files on non UTF-8 Windows system fails due to convert path to str fails
Ad src/capture/mod.rs:113, `new_raw` is accepting a `Option<&str>` as input, this method is not only used by inactive handle, but also used by active handle reading pcap files.

More specific, at src/capture/activated/offline.rs:20, we do `path.as_ref().to_str()`, but apparently this would failed if the path is not a UTF8 encoded Path.

The purpose of `PathBuf` & `Path` is to avoid encoding issue, `to_str()` is not an elegant way.

Hope we could get a quick fix soon :)

cicku:
- pcap_init
- pcap_dump_ftell
- pcap_dump_ftell64
- pcap_dump_file
- pcap_snapshot

---

`pcap_init` fixes #383 and supersedes #384.

Must be merged after #400, and a rebase will be done then.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -302,6 +302,9 @@ pub enum CharEncoding {
 ///
 /// This is optional, but it has to come before any other libpcap call, and a second call asking
 /// for a different encoding fails. Without it strings are in the local character encoding.
+///
+/// On Windows this is what makes a path outside ASCII work, since the local code page is
+/// usually not UTF-8.
 #[cfg(libpcap_1_10_0)]
 pub fn init(encoding: CharEncoding) -> Result<(), Error> {
     Error::with_errbuf(|err| {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -60,8 +60,9 @@
 
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
-use std::ffi::{self, CStr};
+use std::ffi::{self, CStr, CString};
 use std::fmt;
+use std::path::Path;
 
 use self::Error::*;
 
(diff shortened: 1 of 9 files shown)


cicku:
Ready:

Reason of additions:

- `pcap_init`: Non-ASCII handling
- `pcap_dump_ftell64/ftell`:  Size-based file rotation needs 
- `pcap_snapshot`: For `Capture<Offline>` the snapshot length comes from the file header and is otherwise unknowable 

I'm open to drop `pcap_dump_file` if you feel like certain low level functions are not needed. Instead one can always flush() and offset().

src/lib.rs
```diff
@@ -168,6 +172,25 @@ unsafe fn cstr_to_string(ptr: *const libc::c_char) -> Result<Option<String>, Err
...
 
+fn path_to_cstring(path: &Path) -> Result<CString, Error> {
+    #[cfg(not(windows))]
+    let bytes = {
+        use std::os::unix::ffi::OsStrExt;
+        path.as_os_str().as_bytes()
+    };
+    // libpcap has no entry points taking wide strings. It reads the path in the local code page,
+    // or in UTF-8 once pcap_init has been asked for that, so give it the UTF-8 form. A path that
+    // is not valid UTF-8 holds an unpaired surrogate, which has no form libpcap would accept.
```
Wojtek242, line 187: I am not familiar with the term "unpaired surrogate". Is this a standard term in Windows or your term? If it's standard in Windows it can stay. If not, can you rephrase to rephrase this to make it a bit clearer?

Wojtek242 (changes requested):
Looks good. One minor comment to update a comment.

src/lib.rs
cicku: This is Unicode stuff. From U+D800 to U+DFFF are called (1x low + 1x high) surrogates and used for UTF-16. High and low surrogates must be in a pair to present anything higher than 2^16.

https://learn.microsoft.com/en-us/windows/win32/intl/surrogates-and-supplementary-characters

https://www.rfc-editor.org/info/rfc2781/

ortstadtthomas91-coder (approved):

Issue #344, opened 2024-04-15
Use futures Codec for pcap async part


Alexei-Kornienko:
There is quite handy set of traits that can be used with any async stream of data - futures_codec and tokio_util::codec that basically does the same. It would be very handy to use them to work with data received from pcap device. However to do that we need to have some kind of interface that would implement [Sink](https://docs.rs/futures-sink/0.3.29/x86_64-unknown-linux-gnu/futures_sink/trait.Sink.html) and [Stream](https://docs.rs/futures-core/0.3.29/x86_64-unknown-linux-gnu/futures_core/stream/trait.Stream.html).

Is there any chance that such traits can be implemented in this repository?

Wojtek242:
Only if you or somebody else submits a PR. My own activity is limited to maintenance and the occasional feature that piques my interest.

cicku:
Sink makes sense, I can draft a PR.

I do have a question, whether Sink belongs on PacketStream at all, or on a separate PacketSink type?

Wojtek242:
I would prefer a separate PacketSink type given the linked definitions and following the principle of keeping interfaces scoped.

PR #405, opened 2026-08-24
[3.0.0 prep] Async packet sink (sending)

cicku:
Supersedes #373.

### DIFF
--- a/src/sink/mod.rs
+++ b/src/sink/mod.rs
@@ -1,0 +1,407 @@
+//! Support for asynchronous packet transmission.
+//!
+//! See [`Capture::sink`](super::Capture::sink).
+use std::marker::Unpin;
+use std::pin::Pin;
+use std::task::{self, Poll};
+
+use futures::Sink;
+
+#[cfg(target_os = "linux")]
+use {
+    crate::capture::selectable::SelectableCapture, futures::ready, std::io,
+    tokio::io::unix::AsyncFd,
+};
+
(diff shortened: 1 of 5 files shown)


cicku:
This PR is only for sending packets - However, if #344 is looking for a singular solution, we can either:

1. implement Sink on PacketStream where T = Active.
2. make a separate Capture::duplex() returning a PacketDuplex that is both traits.

src/sink/mod.rs
```diff
@@ -0,0 +1,627 @@
...
+            None => return Poll::Ready(Ok(())),
+        };
+
+        // Sending here never waits for the interface, so a sink that is kept fed would never
+        // return Pending and the task it runs in would never let the executor poll anything
+        // else. Two things stop that. The count applies whoever is driving the sink, which off
+        // Linux can be any executor, as it holds nothing of tokio's; 128 is what tokio allows
+        // a task between yields. Spending the task's budget as well holds a task that also
+        // does tokio I/O to one budget between yields rather than one for each source.
+        if self.sent_since_yield == 128 {
```
Wojtek242: Define 128 somewhere as a `const` with a self-explanatory name

Wojtek242 (changes requested):
Tiny comment about a "magic constant"

src/sink/mod.rs
cicku: Done

Issue #393, opened 2025-12-18
Add specific enum value to `pcap::Error` for an interface that has disappeared


GyulyVGC:
Currently if you are capturing packets from an interface and such interface is taken down, the `next_packet` method will indefinitely throw a generic `PcapError("The interface disappeared")`.

If possible, it'd be better to add a new enum value just for this specific error to better identify it in a robust and platform-independent way.
For example this could be useful to break the associated loop.
This is particularly significant since, at least on macOS, such capture will always return that error even in case the corresponding interface is brought back up.

Related issue that would benefit from this: https://github.com/GyulyVGC/sniffnet/issues/1028

Stargateur:
I think we could do better indeed but pcap doesn't really have a lot of rigorous error. pcap mostly define few general error code and than use string as error. 

```c
/*
 * Error codes for the pcap API.
 * These will all be negative, so you can check for the success or
 * failure of a call that returns these codes by checking for a
 * negative value.
 */
#define PCAP_ERROR			-1	/* generic error code */
#define PCAP_ERROR_BREAK		-2	/* loop terminated by pcap_breakloop */
#define PCAP_ERROR_NOT_ACTIVATED	-3	/* the capture needs to be activated */
#define PCAP_ERROR_ACTIVATED		-4	/* the operation can't be performed on already activated captures */
#define PCAP_ERROR_NO_SUCH_DEVICE	-5	/* no such device exists */
#define PCAP_ERROR_RFMON_NOTSUP		-6	/* this device doesn't support rfmon (monitor) mode */
#define PCAP_ERROR_NOT_RFMON		-7	/* operation supported only in monitor mode */
#define PCAP_ERROR_PERM_DENIED		-8	/* no permission to open the device */
#define PCAP_ERROR_IFACE_NOT_UP		-9	/* interface isn't up */
#define PCAP_ERROR_CANTSET_TSTAMP_TYPE	-10	/* this device doesn't support setting the time stamp type */
#define PCAP_ERROR_PROMISC_PERM_DENIED	-11	/* you don't have permission to capture in promiscuous mode */
#define PCAP_ERROR_TSTAMP_PRECISION_NOTSUP -12  /* the requested time stamp precision is not supported */
#define PCAP_ERROR_CAPTURE_NOTSUP	-13	/* capture mechanism not available */

/*
 * Warning codes for the pcap API.
 * These will all be positive and non-zero, so they won't look like
 * errors.
 */
[...]

PR #409, opened 2026-09-08
[3.0.0 prep] Implement libpcap error and warning codes

cicku:
This change originally targeted #393. Amended after https://github.com/the-tcpdump-group/libpcap/commit/ccc5817bd24fd4d6c477507b5f5a0b4194bb0058 goes live in libpcap 1.10.7.


Wojtek242:
LGTM

PR #410, opened 2026-09-08
[3.0.0 prep] Stop Savefile::write reading past the packet data

cicku:
`Savefile::write` no longer reads past the end of the packet data. `caplen` is now capped at the length of the data, and `len` raised to match when it is smaller.

---

Discovered by Rust ASan.


Wojtek242:
LGTM

PR #414, opened 2026-09-13
Build and test on ARM64 Linux and Windows

cicku:
libpcap and npcap have explicitly mentioned ARM64, so we should also test these two.


cicku:
There will be #416 to address dynamically linking issue on Windows, which will turn half of this PR invalid. However, feel free to merge because that one will come out lately.

Wojtek242:
> There will be #416 to address dynamically linking issue on Windows, which will turn half of this PR invalid. However, feel free to merge because that one will come out lately.

What will be easier for you? I'll be waiting with 3.0.0 anyway until you manage to go through all the changes you would like to submit.

cicku:
This one first, then 416. ✅

Wojtek242:
Ok, LGTM

